PR #1, opened 2019-09-18
Modifications to support quantity limits feature

tskimmett:
1. Fix bug where synchronous entity initialization called `set()` asynchronously
2. Fix bug where constant values for properties could be initialized before type dependencies
3. Fix bug where `newValue` was not included in `EntityChangedEventArgs` when rule predicates change in certain cases
4. Add support for serializing a specific property of an entity, optionally forcing an output even if a converter ignores the property.
5. Add support for specifying a custom condition type code for property errors
6. Add support for specifying additional predicates to a `ValidationRule`
7. Add support for additional `ValidationRule` predicates in a property error definition
8. Add support for specifying a resource to use as property error message

### DIFF
--- a/src/entity-serializer.ts
+++ b/src/entity-serializer.ts
@@ -118,6 +118,18 @@ export class EntitySerializer {
 		return injectors;
 	}
 
+	serializePropertyValue(entity: Entity, property: Property, value: any, force = false) {
+		let converters = this._propertyConverters.filter(c => c.shouldConvert(entity, property));
+		if (converters) {
+			for (const converter of converters) {
+				const res = converter.serialize(entity, value, property);
+				if (!force || res !== IgnoreProperty)
+					return res;
+			}
+		}
+		return EntitySerializer.defaultPropertyConverter.serialize(entity, value, property);
+	}
+
 	/**
 	 * Produces a JSON-valid object representation of the entity.
 	 * @param entity
@@ -128,13 +140,7 @@ export class EntitySerializer {
 		flatMap(this.getPropertyInjectors(type), i => i.inject(entity))
 			.concat(type.properties
 				.filter(p => !p.isCalculated && !p.isConstant)
-				.map(prop => {
-					let value = prop.value(entity);
-					let converter = this._propertyConverters.find(c => c.shouldConvert(entity, prop));
-					if (converter)
-						return converter.serialize(entity, value, prop);
-					return EntitySerializer.defaultPropertyConverter.serialize(entity, value, prop);
-				}))
+				.map(prop => this.serializePropertyValue(entity, prop, prop.value(entity))))
 			.forEach(pair => {
 				if (pair && pair !== IgnoreProperty) {
 					if (result.hasOwnProperty(pair.key))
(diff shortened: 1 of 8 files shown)


src/entity-serializer.ts
```diff
@@ -118,6 +118,18 @@ export class EntitySerializer {
 		return injectors;
 	}
 
+	serializePropertyValue(entity: Entity, property: Property, value: any, force = false) {
```
tskimmett: 4

src/entity.ts
```diff
@@ -53,7 +53,7 @@ export class Entity {
 				this.init(properties, context);
 
 			// Raise the initNew or initExisting event on this type and all base types
-			context.ready().then(() => {
+			context.ready(() => {
```
tskimmett: 1

src/rule.ts
```diff
@@ -238,7 +238,7 @@ export class Rule {
 							// Defer change notification until the scope of work has completed
 							EventScope$onExit(() => {
 								rule.returnValues.forEach((returnValue) => {
-									(args.entity.changed as Event<Entity, EntityChangeEventArgs>).publish(args.entity, { entity: args.entity, property: returnValue });
+									(args.entity.changed as Event<Entity, EntityChangeEventArgs>).publish(args.entity, { entity: args.entity, property: returnValue, newValue: returnValue.value(args.entity) });
```
mattheyan: @tskimmett What problem was this causing? I think this is an odd one because the property hasn't actually changed at this point. I think this is a holdover from old code that I'm not sure is relevant anymore (may be with Vue?).

src/entity.ts
```diff
@@ -62,12 +62,9 @@ export class Entity {
 				}
 
 				// Set values of new entity for provided properties
-				if (context.isAsync && isNew && properties)
+				if (isNew && properties)
```
mattheyan: @tskimmett Looks like this `isAsync` property can go away now since its not used? Unless you think there's value in keeping it.

src/rule.ts
tskimmett: I just changed it because typescript complained... Don't know that it was affecting anything.

mattheyan: Was it this line of code that it was complaining about? Or, somewhere else? I'd rather have the `newValue` be optional like `oldValue` (which I thought it was) since because of the way its used its not always possible to provide the value. I don't think its even used though, so I think you could just get rid of `oldValue` and `newValue`. That might actually be a better thing to do anyway.

tskimmett: It was that line. I could see `oldValue` being useful. `newValue` can be obtained by just getting the value of that property for the changed entity., so it's redundant.

mattheyan: Exactly, and that's what subscribers do currently. I agree `oldValue` could be useful in theory, although I don't think anyone is currently using it. This event is kind of funky and honestly I wouldn't expect anyone to use it externally (other than how its used in VueModel). I think it would be fine to do any of the following (option 2 is my favorite):
1) Satisfy the compiler by making `newValue` optional
2) Remove both `oldValue` and `newValue`
3) Remove just `newValue` and keep `oldValue`
4) Leave it as is and I'll circle back to it at some point (I know this wasn't the point of the changes you were making)

tskimmett: I am actually using the entity changed event in application code. Not using old/new value though.

mattheyan: Really, what are you using it for?

PR #2, opened 2019-09-23
ensure ConditionTargets are constructed after obtaining all target properties

tskimmett:
Fix bug where a rule associated with multiple properties would add a condition to an entity where the `ConditionTarget` had an incomplete `properties` list. This caused problems with vuemodel source adapters since they would recompute the `conditions` property before the adapter's property was present in the ConditionTarget's property list.


tskimmett:
Additional note: I don't think we were leveraging the notion of additional `properties` on rules until my changes for quantity limits.

Issue #3, opened 2019-09-25
For some rules, paths are evaluated against the containing type of the property, rather than the target type of the rule


mattheyan:
Example:
```js
{
// ...
BaseType: {
  Foo: String
},
SubType: {
  $extends: "BaseType",
  Index: Number,
  Foo: {
    get: {
      basedOn: "Index",
      get: function() { return "Item " + (this.Index + 1); }
    }
  }
// ...
}
```

The `basedOn` "Index" will fail to resolve the property, since `Index` doesn't exist as a property on `BaseType`.

In `Property.extend()`, there are some references to `this.containingType` which should be `targetType`.

mattheyan:
Fixed by PR #4

PR #4, opened 2019-09-25
Use targetType when resolving rule dependencies

mattheyan:


mattheyan:
Attempt to fix issue #3 
Any reason why the target type should be used in all cases to resolve dependencies?

PR #10, opened 2019-09-27
Add `labelIsFormat` and `helptextIsFormat` getters for properties

mattheyan:
Removes the need for external (ex: app, VueModel) direct access to format class to detect format tokens, so you can instead just work with the property directly


mattheyan:
Going to go ahead and merge this since Taylor has already see the changes, we can revisit if needed...

PR #13, opened 2020-04-30
implement support for resource names as StringFormatRule message

tskimmett:
A resource string can be used as the error message for StringFormatRules as follows:

```
format: {
        reformat: "$&",
        expression: /^([A-Za-z]+)$/,
        message: "string-format-alphabetic"
}
```

### DIFF
--- a/src/string-format-rule.ts
+++ b/src/string-format-rule.ts
@@ -18,14 +18,8 @@ export class StringFormatRule extends ValidationRule {
 			options.name = options.name || "StringFormat";
 
 			// see if the error message is a valid resource: {resource-name}
-			if (typeof options.message === "string" && /^{[^{}]+}$/.test(options.message)) {
-				const resource = options.message.substring(1, options.message.length - 1);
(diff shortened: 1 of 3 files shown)


src/string-format-rule.ts
```diff
@@ -17,14 +17,9 @@ export class StringFormatRule extends ValidationRule {
...
 
-			// ensure the error message is a valid resource
-			if (typeof options.message === "string" && /^{[^{}]+}$/.test(options.message)) {
-				if (rootType.model.getResource(options.message)) {
-					options.message = rootType.model.getResource(options.message);
-				}
-				else {
-					delete options.message;
-				}
+			// see if the error message is a valid resource: {resource-name}
```
mattheyan, line 20: @tskimmett Are the curly braces here a hold-over from the original approach?

tskimmett: @mattheyan Updated this to remove the curly braces.

PR #14, opened 2020-05-19
Standardize Date Time

Nitrokitty:
Standardize the datetime based on the format specifier so that it can be used appropriately for comparisons. When comparing time values, the date will be set to Jan 1 1970. When comparing date values, the time will be set to 12AM.
https://cognitoforms.visualstudio.com/Cognito%20Forms/_workitems/edit/12040/
**Important - Do not complete without reviewing the following PRs:**
- https://github.com/cognitoforms/VueModel/pull/8
- https://cognitoforms.visualstudio.com/Cognito%20Forms/_git/Cognito%20Forms/pullrequest/2339

### DIFF
--- a/src/condition-rule.ts
+++ b/src/condition-rule.ts
@@ -18,6 +18,8 @@ export class ConditionRule extends Rule {
 	// The condition type to raise when asserted
 	conditionType: ConditionType;
 
+	preExecute?: (this: ConditionRule, entity: Entity) => void;
+
 	/**
 	 * Creates a rule that asserts a condition based on a predicate
 	 * @param rootType The model type the rule is for
@@ -32,6 +34,9 @@ export class ConditionRule extends Rule {
 	
 		// assertion function
 		this.assert = options.assert;
+
+		// pre-execute
+		this.preExecute = options.preExecute;
 
 		// message
 		this.message = options.message;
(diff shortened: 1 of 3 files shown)


src/condition-rule.ts
```diff
@@ -51,6 +56,10 @@ export class ConditionRule extends Rule {
 		let assert: boolean;
 		let message: string;
 
+		if(this.preExecute) {
```
Nitrokitty, line 59: Add the ability to do some setup before executing the rule

src/range-rule.ts
```diff
@@ -63,18 +47,94 @@ export class RangeRule extends ValidationRule {
 			else
 				return rootType.model.getResource("range-at-most").replace("{max}", Property$format(options.property, range.max) || range.max);
 		};
+		
+		options.preExecute = (entity) => {
```
Nitrokitty, line 51: Before executing the range rule, store the new range on the RangeRule entity.

src/range-rule.ts
```diff
@@ -63,18 +47,94 @@ export class RangeRule extends ValidationRule {
...
 		};
+		
+		options.preExecute = (entity) => {			
+			this["range"] = RangeRule.getRange(entity, options);
+		}
 
 		// call the base type constructor
 		super(rootType, options);
+
+		Object.defineProperty(this, "range", { value: null, writable: true });
```
Nitrokitty, line 58: Defined like so to avoid having to move the `super` call to the first line of the constructor

tskimmett:
I think I'm good with the change, the only request I would make is not using static class methods. So instead of Standardize.standardize it would be nice if there was just a standardize()  function or standardizeDate() function.

Nitrokitty:
> I think I'm good with the change, the only request I would make is not using static class methods. So instead of Standardize.standardize it would be nice if there was just a standardize() function or standardizeDate() function.

Done! :D

PR #20, opened 2020-06-16
feat: asynchronous Entity.set when calling Type.create

tskimmett:
Because `Type.create()` may update the pooled entity using `set()`, if the provided state includes values that must be asynchronously resolved, an error would be thrown because async value resolution was only supported by `Entity.init()`. This change introduces a way to provide an `InitializationContext` and perform an action on an `Entity`, allowing us to effectively support async `set()`.

```ts
entity.withContext(context, e => e.set(state));
```

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -160,14 +160,11 @@ export class Entity {
 		for (let [propName, state] of Entity.getSortedPropertyData(properties)) {
 			const prop = this.serializer.resolveProperty(this, propName);
 			if (prop) {
-				if (this._context) {
-					const valueResolution = this._context.tryResolveValue(this, prop, state);
-					if (valueResolution) {
-						valueResolution.then(asyncState => this.setProp(prop, asyncState));
-						continue;
-					}
-				}
-				this.setProp(prop, state);
+				const valueResolution = this._context ? this._context.tryResolveValue(this, prop, state) : null;
+				if (valueResolution)
+					valueResolution.then(asyncState => this.setProp(prop, asyncState));
+				else
+					this.setProp(prop, state);
 			}
 		}
 	}
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -184,21 +184,24 @@ export class Entity {
 						// If the item is an object that has an Id property, then retrieve or create an object with that Id
 						if (!(s instanceof ChildEntity) && typeof s === "object" && getIdFromState(ChildEntity.meta, s)) {
 							const id = getIdFromState(ChildEntity.meta, s);
-							if (id && ChildEntity.meta.get(id))
-								ChildEntity.meta.get(id).withContext(this._context, entity => entity.set(state));
+							if (id && ChildEntity.meta.get(id)) {
+								ChildEntity.meta.get(id).withContext(this._context, entity => entity.set(s));
+								s = null;
+							}
 							else
 								s = new (ChildEntity as any)(id, s, this._context);
 						}
+
 						if (s instanceof ChildEntity)
 							state.splice(idx, 1, s);
-						else
+						else if (s)
 							currentValue[idx].withContext(this._context, entity => entity.set(s));
 					}
 					// Add a list item
 					else if (s instanceof ChildEntity)
 						currentValue.push(s);
 					else
-						currentValue.push(ChildEntity.meta.createSync(s));
+						currentValue.push(new (ChildEntity as any)(s, this._context));
 				});
 			}
 			else if (state instanceof ChildEntity)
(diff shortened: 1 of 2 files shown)


src/entity.ts
```diff
@@ -130,86 +128,119 @@ export class Entity {
 		Property$init(prop, this, value);
 	}
 
+	withContext(context: InitializationContext, action: (entity: Entity) => void) {
```
mattheyan, line 131: @tskimmett I think the code below makes sense. Just curious though, what's the scenario where `withContext` gets called and there's already a context?

src/entity.ts
```diff
@@ -130,86 +128,119 @@ export class Entity {
...
-								else
-									currentValue[idx].set(s);
-							}
-							// Add a list item
-							else if (s instanceof ChildEntity)
-								currentValue.push(s);
-							else
-								currentValue.push(ChildEntity.meta.createSync(s));
-						});
+				if (this._context) {
```
mattheyan, line 163: What do you think about getting rid of the nested `if` and `continue` and making it an `if`/`else`, using a ternary (like `valueResolution = this._context ? this._context.tryResolveValue(this, prop, state) : null`)? Could be just my personal preference, but I think code that uses `break`/`continue` can be a bit harder to understand, and may be more susceptible to refactor bugs. I don't have any data to back that up, just a hunch.

src/entity.ts
```diff
@@ -130,86 +128,119 @@ export class Entity {
...
+					if (!(s instanceof ChildEntity))
+						s = this.serializer.deserialize(this, s, prop, this._context, false);
+
+					// Modifying/replacing existing list item
+					if (idx < currentValue.length) {
+						// If the item is an object that has an Id property, then retrieve or create an object with that Id
+						if (!(s instanceof ChildEntity) && typeof s === "object" && getIdFromState(ChildEntity.meta, s)) {
+							const id = getIdFromState(ChildEntity.meta, s);
+							if (id && ChildEntity.meta.get(id))
+								ChildEntity.meta.get(id).withContext(this._context, entity => entity.set(state));
```
mattheyan, line 191: If we don't assign `s` here, does that result in `set` being called twice?

src/entity.ts
```diff
@@ -130,86 +128,119 @@ export class Entity {
...
+							else
+								s = new (ChildEntity as any)(s, this._context);
```
mattheyan, line 193: If the state has an id but the instance doesn't exist yet, does it need to pass the id to the constructor?

src/entity.ts
```diff
@@ -130,86 +128,119 @@ export class Entity {
...
+			const ChildEntity = prop.propertyType;
+			if (prop.isList && Array.isArray(state) && Array.isArray(currentValue)) {
+				state.forEach((s, idx) => {
+					if (!(s instanceof ChildEntity))
+						s = this.serializer.deserialize(this, s, prop, this._context, false);
+
+					// Modifying/replacing existing list item
+					if (idx < currentValue.length) {
+						// If the item is an object that has an Id property, then retrieve or create an object with that Id
+						if (!(s instanceof ChildEntity) && typeof s === "object" && getIdFromState(ChildEntity.meta, s)) {
```
mattheyan, line 188: When I saw this I wondered if there was a way we could call `create` instead. But, I looked at `create` and see that it takes a value resolver, so that would make the method signature more complex. But, maybe its worth considering if we can avoid duplicating the code? Just leaving this here to start the discussion...

mattheyan: I see we also have a similar situation below in the non-list scenario.

src/entity.ts
tskimmett: There isn't a known scenario, but I figured it should be accounted for.

src/entity.ts
tskimmett: Yeah, I didn't have a great feeling about the use of `continue` here. I'll rework that.

src/entity.ts
tskimmett: Yes.

src/entity.ts
mattheyan: Makes sense 👍

src/entity.ts
tskimmett: Fixed.

src/entity.ts
tskimmett: Fixed.

src/entity.ts
tskimmett: Fixed.

src/entity.ts
tskimmett: It runs into the same problem as modifying the `set()` signature, they are both publicly exposed, so adding a parameter would alter the public api.

mattheyan:
@tskimmett I thought you had a few more pending changes to commit?

PR #21, opened 2020-06-18
fix: bug with nested async value resolutions during createOrUpdate

tskimmett:
```
{ Lookup: "1-1" }

1-1 => { 
   NestedLookup: "2-1"
}

// this entity would not be initialized correctly because the context had been cleared already
2-1 => { another entity }
```

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -160,14 +160,11 @@ export class Entity {
 		for (let [propName, state] of Entity.getSortedPropertyData(properties)) {
 			const prop = this.serializer.resolveProperty(this, propName);
 			if (prop) {
-				if (this._context) {
-					const valueResolution = this._context.tryResolveValue(this, prop, state);
-					if (valueResolution) {
-						valueResolution.then(asyncState => this.setProp(prop, asyncState));
-						continue;
-					}
-				}
-				this.setProp(prop, state);
+				const valueResolution = this._context ? this._context.tryResolveValue(this, prop, state) : null;
+				if (valueResolution)
+					valueResolution.then(asyncState => this.setProp(prop, asyncState));
+				else
+					this.setProp(prop, state);
 			}
 		}
 	}
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -184,21 +184,24 @@ export class Entity {
 						// If the item is an object that has an Id property, then retrieve or create an object with that Id
 						if (!(s instanceof ChildEntity) && typeof s === "object" && getIdFromState(ChildEntity.meta, s)) {
 							const id = getIdFromState(ChildEntity.meta, s);
-							if (id && ChildEntity.meta.get(id))
-								ChildEntity.meta.get(id).withContext(this._context, entity => entity.set(state));
+							if (id && ChildEntity.meta.get(id)) {
+								ChildEntity.meta.get(id).withContext(this._context, entity => entity.set(s));
+								s = null;
+							}
 							else
 								s = new (ChildEntity as any)(id, s, this._context);
 						}
+
 						if (s instanceof ChildEntity)
 							state.splice(idx, 1, s);
-						else
+						else if (s)
 							currentValue[idx].withContext(this._context, entity => entity.set(s));
 					}
 					// Add a list item
 					else if (s instanceof ChildEntity)
 						currentValue.push(s);
 					else
-						currentValue.push(ChildEntity.meta.createSync(s));
+						currentValue.push(new (ChildEntity as any)(s, this._context));
 				});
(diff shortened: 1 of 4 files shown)


src/entity.ts
```diff
@@ -134,7 +134,7 @@ export class Entity {
 		if (!this._context)
 			this._context = context;
 		// Ensure provided context waits on the existing context to be ready
-		else
+		else if (this._context !== context)
```
tskimmett, line 137: There was a sneaky bug where revisiting an entity would result in a deadlock (the context would be waiting on itself to be ready).

src/initilization-context.ts
```diff
@@ -18,8 +18,12 @@ export class InitializationContext {
 		this.tasks.add(task);
 		task.then(() => {
 			this.tasks.delete(task);
-			if (this.tasks.size === 0)
-				this.waiting.forEach(done => done());
+			// allow additional tasks to be queued as a result of this one
+			Promise.resolve().then(() => {
```
tskimmett, line 22: Wait on an immediately resolved promise to make sure we check the `tasks` list _after_ allowing other tasks to be queued.

src/entity.ts
mattheyan: Makes sense

mattheyan, line 137: Are there any implications on the `if (context !== null) { ...` code at the bottom of the method? Previously it would either set the context and then null it out, or it would wait on the given context, then null it out, so the nulling out was at worst redundant in the second case? I suppose the same is still true? Still, might be more clear to only wait and null out the context in the method call when its assigned?

src/initilization-context.ts
mattheyan: I understand in the abstract sense that async init code could queue other async init code, but its difficult to make out the precise scenario just looking at the code. Just to make sure I understand... A `wait()` task was queued to async resolve a value via `tryResolveValue` (ex: a lookup entity), and when it completed we would fire off the queue callbacks (the `waiting` list) - which would resolve the context created for "create" calls and/or constructor calls. But, if the queued task (i.e. the `tryResolveValue` call) required additional waits on the context due to nested calls to constructors or "create" (via a "set" call after value resolution?), then we want it to wait for those to get queued up (the context's promise callback may fire before the value resolver makes a recursive call to `wait`?), hence the immediately resolving promise.

mattheyan: Also, its a minor thing, but I think the `done` naming used in the previous version made it a bit more clear. And the use of `shift()()` is harder to grok.

src/initilization-context.ts
```diff
@@ -18,8 +18,12 @@ export class InitializationContext {
...
+				if (this.tasks.size === 0)
+					while (this.waiting.length > 0)
+						this.waiting.shift()();
+			});
 		});
 	}
```
mattheyan, line 29: Should the `waiting.push` below not happen if the task size is 0? Not sure if it matters, but it at least looks like a lurking bug.

src/initilization-context.ts
mattheyan: Also, just a thought: what do you think about a new helper method for this (ex: `nextTick`)

src/entity.ts
```diff
@@ -134,7 +134,7 @@ export class Entity {
...
 			context.wait(new Promise(resolve => this._context.ready(resolve)));
```
mattheyan, line 138: This is obviously outside of the scope of these changes, but I wonder if this code might read better with something like `context.wait(this._context.asPromise())`, or `context.waitForContext(this._context)`, or some other variation? Or, `ready` could return a promise so it could be called like `context.wait(this._context.ready())`? Or, the context itself could be promise-like (essentially `ready` replaced with `then`) so that it could be called like `context.wait(this._context)`? Anyway, just throwing the idea out there for consideration since we're looking at this code at the moment.

src/initilization-context.ts
tskimmett: Your understanding is correct. We shouldn't try to run the `ready()` callbacks until the caller of `tryResolveValue` has a chance to process the result.

I will update the `shift()()` code to be more clear.

tskimmett: Updated

src/entity.ts
tskimmett: Fixed.

src/initilization-context.ts
tskimmett: Fixed

src/entity.ts
tskimmett: Implemented this idea.

PR #22, opened 2020-07-21
Rule Prioritization

Nitrokitty:
Pass through the rule or format error through to the condition to be saved on the condition type as `source`

Must be merged at the same time with the following PRs:
- https://cognitoforms.visualstudio.com/Cognito%20Forms/_git/Cognito%20Forms/pullrequest/2631
- https://github.com/cognitoforms/VueModel/pull/12

### DIFF
--- a/src/condition-rule.ts
+++ b/src/condition-rule.ts
@@ -40,7 +40,7 @@ export class ConditionRule extends Rule {
 		this.conditionType = options.conditionType ? (typeof options.conditionType === "string" ? 
 			ConditionType.get(options.conditionType) : 
 			options.conditionType) :
-			Rule$ensureConditionType(options.name, rootType, options.category || "ErrorConditionType");
+			Rule$ensureConditionType(this, options.name, rootType, options.category || "ErrorConditionType");		
 
 		// properties
 		this.properties = options.properties;
--- a/src/condition-rule.ts
+++ b/src/condition-rule.ts
@@ -40,7 +40,7 @@ export class ConditionRule extends Rule {
 		this.conditionType = options.conditionType ? (typeof options.conditionType === "string" ? 
 			ConditionType.get(options.conditionType) : 
 			options.conditionType) :
-			Rule$ensureConditionType(this, options.name, rootType, options.category || "ErrorConditionType");		
+			Rule$ensureConditionType(this, options.name, rootType, options.category || "Error");		
 
 		// properties
 		this.properties = options.properties;
--- a/src/condition-rule.ts
+++ b/src/condition-rule.ts
@@ -63,7 +63,9 @@ export class ConditionRule extends Rule {
 
 		// create or remove the condition if necessary
 		if (typeof assert !== "undefined") {
-			this.conditionType.when(assert, entity, this.properties, message);
+			// pass through "this" so it can add the source when the rule is constructed
+			// Add the source to the condition instead of the condition type
+			this.conditionType.when(this, assert, entity, message);
 		}
 	}
 
--- a/src/condition-rule.ts
+++ b/src/condition-rule.ts
@@ -40,7 +40,7 @@ export class ConditionRule extends Rule {
 		this.conditionType = options.conditionType ? (typeof options.conditionType === "string" ? 
 			ConditionType.get(options.conditionType) : 
 			options.conditionType) :
-			Rule$ensureConditionType(this, options.name, rootType, options.category || "Error");		
+			Rule$ensureConditionType(options.name, rootType, options.category || "Error");
 
 		// properties
 		this.properties = options.properties;
--- a/src/condition.ts
+++ b/src/condition.ts
@@ -5,14 +5,11 @@ import { ConditionTarget } from "./condition-target";
 import { PropertyPath } from "./property-path";
 import { ObservableArray } from "./observable-array";
 import { Property } from "./property";
-import { Rule } from "./rule";
 
 export class Condition {
 	type: ConditionType;
 	message: string;
-	origin: string;
 	targets: ObservableArray<ConditionTarget>;
-	source: Rule | FormatError;
 
 	/**
 		* Creates a condition of a specific type associated with one or more entities in a model.
@@ -21,10 +18,9 @@ export class Condition {
 		* @param target The root target entity the condition is associated with.
 		* @param properties The set of property paths specifying which properties and entities the condition should be attached to.
 		*/
-	constructor(type: ConditionType, message: string, target: Entity, source: Rule | FormatError, properties: PropertyPath[] = []) {
+	constructor(type: ConditionType, message: string, target: Entity, properties: PropertyPath[] = []) {
 		this.type = type;
-		this.message = message || (type ? type.message : undefined);
-
+		this.message = message || (type ? type.message : undefined);		
 		let targets = this.targets = ObservableArray.create<ConditionTarget>();
 
 		// create targets if a root was specified
@@ -125,7 +121,7 @@ export class Condition {
 }
 
 export interface ConditionConstructor {
-	new(type: ConditionType, message: string, target: Entity, properties: string[], origin?: string): Condition;
+	new(type: ConditionType, message: string, target: Entity, properties: string[]): Condition;
 }
 
 export interface ConditionsChangedEventArgs {
--- a/src/condition.ts
+++ b/src/condition.ts
@@ -5,12 +5,14 @@ import { ConditionTarget } from "./condition-target";
 import { PropertyPath } from "./property-path";
 import { ObservableArray } from "./observable-array";
 import { Property } from "./property";
+import { Rule } from "./rule";
 
 export class Condition {
 	type: ConditionType;
 	message: string;
 	origin: string;
 	targets: ObservableArray<ConditionTarget>;
+	source: Rule | FormatError;
 
 	/**
 		* Creates a condition of a specific type associated with one or more entities in a model.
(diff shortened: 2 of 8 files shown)


src/format-error.ts
```diff
@@ -11,7 +11,7 @@ export class FormatError {
 
 	constructor(message: string, invalidValue: any) {
 		if (FormatError.ConditionType === null) {
-			FormatError.ConditionType = new ErrorConditionType("FormatError", "The value is not properly formatted.");
+			FormatError.ConditionType = new ErrorConditionType("FormatError", "The value is not properly formatted.", this);
```
tskimmett, line 14: Passing `this` to the `ErrorConditionType` constructor here may not be correct, given that it appears to be some type of static initialization. @mattheyan are you familiar with this at all? In my view, "correctness" would hinge on how the condition type's `source` ends up being used. I.e. if downstream we reference something that would be the same for every format error, it would be fine, though that is not guaranteed by any means.

mattheyan: Yeah, I think `FormatError` is basically just a wrapper around the condition. Seems like the more appropriate source would be the `Format` object.

src/property.ts
```diff
@@ -443,7 +443,7 @@ export class Property implements PropertyPath {
 
 					let conditionType: ConditionType;
 					if (errorOptions.code)
-						conditionType = ConditionType.get(errorOptions.code) || new ErrorConditionType(errorOptions.code, "error");
+						conditionType = ConditionType.get(errorOptions.code) || new ErrorConditionType(errorOptions.code, "error", null);
```
mattheyan, line 446: So b/c we're creating the condition type before constructing the rule, it won't have a source. I think the rule class will get or create its condition type, so I'm thinking we may be able to simplify this to `conditionType = errorOptions.code`?

mattheyan: Well, I'm not sure the logic is quite the same actually. See `ConditionRule` constructor:
```
		this.conditionType = options.conditionType ? (typeof options.conditionType === "string" ? 
			ConditionType.get(options.conditionType) : 
			options.conditionType) :
			Rule$ensureConditionType(options.name, rootType, options.category || "Error")
```

src/format-error.ts
tskimmett: I think I was getting more at the fact we are providing a potentially random instance to a static object. Is that not a problem?

mattheyan: Oh, I see. The condition type on format error is static. I missed that detail. Yeah, I agree.

src/condition-rule.ts
```diff
@@ -40,7 +40,7 @@ export class ConditionRule extends Rule {
 		this.conditionType = options.conditionType ? (typeof options.conditionType === "string" ? 
 			ConditionType.get(options.conditionType) : 
 			options.conditionType) :
-			Rule$ensureConditionType(options.name, rootType, options.category || "ErrorConditionType");
+			Rule$ensureConditionType(this, options.name, rootType, options.category || "ErrorConditionType");
```
mattheyan, line 43: This is not due to your changes, but one of the bugs I identified when planning this was that "ErrorConditionType" should be "Error". Do you mind fixing that while you're in here?

src/condition.ts
```diff
@@ -5,12 +5,14 @@ import { ConditionTarget } from "./condition-target";
 import { PropertyPath } from "./property-path";
 import { ObservableArray } from "./observable-array";
 import { Property } from "./property";
+import { Rule } from "./rule";
 
 export class Condition {
 	type: ConditionType;
 	message: string;
 	origin: string;
```
mattheyan, line 13: Another pre-existing issue is that `origin` does not appear to be used

src/rule.ts
```diff
@@ -344,7 +344,7 @@ function executeRule(rule: Rule, obj: Entity): void {
 	});
 };
 
-export function Rule$ensureConditionType<DesiredConditionType = ErrorConditionType | WarningConditionType>(ruleName: string, typeOrProp: Type | Property, category: string = "Error"): ErrorConditionType | WarningConditionType {
+export function Rule$ensureConditionType<DesiredConditionType = ErrorConditionType | WarningConditionType>(rule: Rule, ruleName: string, typeOrProp: Type | Property, category: string = "Error"): ErrorConditionType | WarningConditionType {
```
mattheyan, line 347: One final pre-existing issue, the `<DesiredConditionType = ErrorConditionType | WarningConditionType>` generic type parameter doesn't make sense and is not needed.

mattheyan, line 347: Also, the default category of "Error" doesn't seem appropriate.

src/format-error.ts
mattheyan: Maybe the solution is to not pass in the source to condition type (allow it to be null if that's not possible already), but pass in the `Format` instead in `createCondition` below?

mattheyan: Of course, if we did that we'd have to check the condition's source, not the condition type.

src/condition.ts
```diff
@@ -19,7 +21,7 @@ export class Condition {
 		* @param target The root target entity the condition is associated with.
 		* @param properties The set of property paths specifying which properties and entities the condition should be attached to.
 		*/
-	constructor(type: ConditionType, message: string, target: Entity, properties: PropertyPath[] = []) {
+	constructor(type: ConditionType, message: string, target: Entity, source: Rule | FormatError, properties: PropertyPath[] = []) {
```
mattheyan, line 24: The source argument isn't assigned in the constructor

src/property.ts
Nitrokitty: so this should be changed to `conditionType = errorOptions.code` and then it will use the code when constructing the condition type later on (in the block of code you posted)?

src/condition.ts
Nitrokitty: fixed.... that's an odd one. i think i started this before i got sick and completely forgot about it lol

src/rule.ts
Nitrokitty: fixed!

Nitrokitty: fixed!

src/condition.ts
Nitrokitty: fixed!

src/condition-rule.ts
Nitrokitty: fixed!

src/property.ts
mattheyan: Sorry for taking so long to get back to this, but I think our most recent change of plans negates this, since we'll be storing the source on the condition rather than the condition type.

src/format-error.ts
```diff
@@ -2,25 +2,29 @@ import { ErrorConditionType } from "./condition-type";
...
 import { Property } from "./property";
 import { Condition } from "./condition";
+import { Format } from "./format";
 
-export class FormatError {
+export class FormatError {	
 	readonly messageTemplate: string;
 	readonly invalidValue: any;
-
+	source: Format<any>;
```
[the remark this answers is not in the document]
tskimmett: +1 for `format`

src/rule.ts
```diff
@@ -344,7 +344,7 @@ function executeRule(rule: Rule, obj: Entity): void {
 	});
 };
 
-export function Rule$ensureConditionType<DesiredConditionType = ErrorConditionType | WarningConditionType>(ruleName: string, typeOrProp: Type | Property, category: string = "Error"): ErrorConditionType | WarningConditionType {
+export function Rule$ensureConditionType(rule: Rule, ruleName: string, typeOrProp: Type | Property, category: string): ErrorConditionType | WarningConditionType {
```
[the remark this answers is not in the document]
Nitrokitty: Fixed!

src/format-error.ts
```diff
@@ -2,25 +2,29 @@ import { ErrorConditionType } from "./condition-type";
 import { Entity } from "./entity";
 import { Property } from "./property";
 import { Condition } from "./condition";
+import { Format } from "./format";
 
-export class FormatError {
+export class FormatError {
```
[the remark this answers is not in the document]
Nitrokitty: Fixed! i do but I don't think I have it enabled for this repo... I think it keeps asking me to enable it and I keep clicking "not now" lol! Will enable it :D

src/format-error.ts
```diff
@@ -2,25 +2,29 @@ import { ErrorConditionType } from "./condition-type";
...
 
+		this.source = source;
 		this.messageTemplate = message;
 		this.invalidValue = invalidValue;
 	}
 
 	createCondition(target: Entity, prop: Property): Condition {
 		// TODO: Format error doesn't handle tokens?
-		return new Condition(FormatError.ConditionType, this.messageTemplate.replace("{property}", prop.label), target, [prop]);
+		// pass through "this"
```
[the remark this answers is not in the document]
Nitrokitty: Fixed!

src/condition-rule.ts
```diff
@@ -40,7 +40,7 @@ export class ConditionRule extends Rule {
 		this.conditionType = options.conditionType ? (typeof options.conditionType === "string" ? 
 			ConditionType.get(options.conditionType) : 
 			options.conditionType) :
-			Rule$ensureConditionType(options.name, rootType, options.category || "ErrorConditionType");
+			Rule$ensureConditionType(this, options.name, rootType, options.category || "Error");
```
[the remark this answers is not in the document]
Nitrokitty: Fixed!

src/property.ts
mattheyan: I believe the added `null` should be removed. I think it was for the new `source` argument, and now is falling back to mapping to the optional `sets` argument.

Nitrokitty: Fixed!

PR #23, opened 2020-07-22
Make sure the date and number exists before formatting

Nitrokitty:
Work Item: [Calculated range causes time fields to be conditionally visible](https://cognitoforms.visualstudio.com/Cognito%20Forms/_workitems/edit/12039/)

### DIFF
--- a/src/globalization.ts
+++ b/src/globalization.ts
@@ -514,6 +514,9 @@ function padYear(year: number): string {
  * @param cultureInfo The culture
  */
 export function formatDate(date: Date, format: string, cultureInfo: CultureInfo): string {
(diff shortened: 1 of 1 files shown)


Nitrokitty:
Yes! The issues is that if the date is null, then when we call `date.getHours()` in the foreach loop's switches we get an error. 
In the example, it occurs when we're trying to format the `start` value for the dateTime field `"start":$format(fieldSlot.field.lastTarget.Time_Minimum, "H:mm")`
Time_Minimum's value is null at time of render causing it to fail on `case "H":` (`ret.append(date.getHours().toString());)`

PR #24, opened 2020-07-27
fix: remove references to remove/removeAt functions

tskimmett:
These references to `remove` and `removeAt` look to be carry over from ExoWeb, and one of the if clauses was causing an error due to checking for the existence of the `remove` function and then trying to call `removeAt`.

### DIFF
--- a/src/observable-array.ts
+++ b/src/observable-array.ts
@@ -566,15 +566,7 @@ function observableSplice(arr: any[], events: any[], index: number, removeCount:
 	let arr2 = arr as any;
 
 	if (removeCount) {
-		if (removeCount > 1 && arr2.removeRange) {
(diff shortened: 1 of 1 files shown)


src/observable-array.ts
```diff
@@ -566,15 +566,7 @@ function observableSplice(arr: any[], events: any[], index: number, removeCount:
 	let arr2 = arr as any;
 
 	if (removeCount) {
-		if (removeCount > 1 && arr2.removeRange) {
-			removedItems = arr2.removeRange(index, removeCount);
-		}
-		else if (removeCount === 1 && arr2.remove) {
-			removedItems = [arr2.removeAt(index)];
```
tskimmett, line 573: This line was throwing an error, since it calls `removeAt` after checking the existence of `remove`.

PR #27, opened 2020-08-12
fix bugs with asynchronous updating of entities

tskimmett:
## Specific bugs
- infinite recursion when an entity graph is circular via asynchronously resolved properties
- premature execution of waiting callbacks on the initialization context preventing async resolution of some property values

## Description of changes
- refactor `Entity.withContext` to be `Entity.updateWithContext`, making it serve the specific purpose of calling `set()` with a given `InitializationContext` (this lets us be more rigorous in the implementation, specifically allowing the reentrancy clause)
- prevent reentrant calls to `Entity.updateWithContext` for the same entity with the same context (avoids infinite recursion)
- short circuit processing of `waiting` queue on `InitializationContext` if a waiting callback queues a new task on the context
- tweak logic around updating entity lists in `Entity.setProp` to update the list item if
    + its identifier matches the identifier in the state object or
    + the entities in the list are not pooled (no identifier)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   "scripts": {
     "type-check": "tsc --noEmit",
     "type-check:watch": "npm run type-check -- --watch",
-    "build:typescript": "rimraf ./@types && rimraf ./lib && tsc",
-    "build:webpack": "rimraf ./out && webpack --mode=production",
+    "build:typescript": "tsc",
+    "build:webpack": "webpack --mode=production",
     "build": "npm run build:typescript && npm run build:webpack",
     "build:watch": "nodemon -e ts --watch src/ -x \"npm run build\"",
     "test": "jest",
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -71,7 +71,7 @@ export class Entity {
 
 				// Set values of new entity for provided properties
 				if (isNew && properties)
-					this.withContext(context, () => this.set(properties));
+					this.updateWithContext(context, properties);
 			});
 		}
 	}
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -9,8 +9,8 @@
   "scripts": {
     "type-check": "tsc --noEmit",
     "type-check:watch": "npm run type-check -- --watch",
-    "build:typescript": "rimraf ./@types && rimraf ./lib && tsc",
-    "build:webpack": "rimraf ./out && webpack --mode=production",
+    "build:typescript": "tsc",
```
tskimmett, line 12: Getting rid of `rimraf` calls makes npm linking in the product repo smoother, because webpack does not crash as a result of dependency files being deleted.

tskimmett: This means you can run `npm run build:watch` in model.js, make changes, and things will be automatically recompiled all the way down to the product dev server.

mattheyan: Nice! I suspect I was using `rimraf` to account for deleted or renamed source files, which would have been much more likely to happen during early development, but not so much anymore.

src/entity.ts
```diff
@@ -218,7 +227,7 @@ export class Entity {
 				else if (typeof state !== "object")
 					value = state;
 				else if (currentValue && !getIdFromState(ChildEntity.meta, state))
-					currentValue.withContext(this._context, entity => entity.set(state));
+					currentValue.updateInContext(this._context, state);
```
mattheyan, line 230: Is this a typo?

tskimmett: Yep. Intellisense must not have auto-renamed that occurrence. Thanks.

PR #28, opened 2020-08-18
Change normalize to accept a format string

mattheyan:
The method was checking for constants "d" and "t" anyway, so its functionally equivalent to just take the string arg directly. This probably should be an instance method on the model, since formats depend on localization, but since it doesn't make a difference in the way its used I propose this simple change for now.


tskimmett:
@mattheyan Can you update the changelog to reflect this change? Other than that everything looks good, and definitely appreciate the tests.

PR #29, opened 2020-08-28
fix timing issue with circular calculation and stackoverflow

tskimmett:
This bug demonstrates the calculation issue: https://dev.azure.com/cognitoforms/Cognito%20Forms/_workitems/edit/16916

Setup
---
A form with
- a repeating section **RS**
  - with a number field **Num**
- a section **S**
  - with a calculation **C** based on the sum of **Form.RS.Num** <-- the circular nature of `Form` seems to be key here
    - and a custom error that references **C**

Behavior
---
When initializing an existing entry that contains multiple items in **RS**, **C** is not initialized correctly because it seems the custom error triggers initialization of **C** _before_ **RS** has been deserialized/populated. In short, **S** is "fully initialized" before we have even processed/initialized the state of **RS**.

Fix
---
I've added a synchronous blocking mechanism to the `InitializationContext` which prevents any part of the object graph from "fully initializing" (init events, etc.) until the rest of the state graph has been processed.

Other changes
---
- fixed a bug with calculated list properties which causes stack overflow

Testing
---
- added a test to cover the calculated list property stackoverflow
- I couldn't create a negative test case manually, but I've added a test to cover this issue in the product codebase

### DIFF
--- a/src/calculated-property-rule.ts
+++ b/src/calculated-property-rule.ts
@@ -1,6 +1,6 @@
 import { Rule, RuleInvocationOptions, RuleOptions } from "./rule";
 import { Type } from "./type";
-import { Property, PropertyRuleOptions } from "./property";
+import { Property, PropertyRuleOptions, Property$init } from "./property";
 import { Entity } from "./entity";
 import { ObservableArray, updateArray } from "./observable-array";
 import { RuleInvocationType } from "./rule-invocation-type";
@@ -108,31 +108,36 @@ export class CalculatedPropertyRule extends Rule {
 
 		// modify list properties to match the calculated value instead of overwriting the property
 		if (this.property.isList) {
-			// re-calculate the list values
-			var newList = newValue;
+			if (!this.property.isInited(obj)) {
+				Property$init(this.property, obj, newValue);
+			}
+			else {
+				// re-calculate the list values
+				var newList = newValue;
 
-			// compare the new list to the old one to see if changes were made
-			var curList = this.property.value(obj) as ObservableArray<any>;
+				// compare the new list to the old one to see if changes were made
+				var curList = this.property.value(obj) as ObservableArray<any>;
 
-			if (newList.length === curList.length) {
-				var noChanges = true;
+				if (newList.length === curList.length) {
+					var noChanges = true;
 
-				for (var i = 0; i < newList.length; ++i) {
-					if (newList[i] !== curList[i]) {
-						noChanges = false;
-						break;
+					for (var i = 0; i < newList.length; ++i) {
+						if (newList[i] !== curList[i]) {
+							noChanges = false;
+							break;
+						}
+					}
+
+					if (noChanges) {
+						return;
 					}
 				}
 
-				if (noChanges) {
-					return;
-				}
+				// update the current list so observers will receive the change events
+				curList.batchUpdate((array) => {
+					updateArray(array, newList);
+				});
 			}
-
-			// update the current list so observers will receive the change events
-			curList.batchUpdate((array) => {
-				updateArray(array, newList);
-			});
 		}
 		else {
 			// Otherwise, just set the property to the new value
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -143,7 +143,7 @@ export class Entity {
 			this._context = context;
 		// Ensure provided context waits on the existing context to be ready
 		else if (this._context !== context)
-			context.wait(this._context.readyPromise);
+			context.wait(this._context.ready);
 
 		this.set(state);
 
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -57,11 +57,15 @@ export class Entity {
 			type.register(this);
 
 			// Initialize existing entity with provided property values
-			if (!isNew && properties)
+			if (!isNew && properties) {
+				// We need to pause processing of callbacks to prevent publishing entity events while still processing the state graph
+				const resumeContextQueue = context.delayQueue();
 				this.init(properties, context);
+				resumeContextQueue();
+			}
 
 			// Raise the initNew or initExisting event on this type and all base types
-			context.ready(() => {
+			context.whenReady(() => {
 				for (let t = type; t; t = t.baseType) {
 					if (isNew)
 						(t.initNew as Event<Type, EntityInitNewEventArgs>).publish(t, { entity: this });
@@ -144,7 +148,7 @@ export class Entity {
 		this.set(state);
 
 		if (context !== null && !hadContext) {
-			context.ready(() => {
+			context.whenReady(() => {
 				this._context = null;
 			});
 		}
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -59,9 +59,7 @@ export class Entity {
 			// Initialize existing entity with provided property values
 			if (!isNew && properties) {
 				// We need to pause processing of callbacks to prevent publishing entity events while still processing the state graph
-				const resumeContextQueue = context.delayQueue();
-				this.init(properties, context);
-				resumeContextQueue();
+				context.execute(() => this.init(properties, context));
 			}
 
 			// Raise the initNew or initExisting event on this type and all base types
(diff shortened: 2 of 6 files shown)


src/calculated-property-rule.ts
```diff
@@ -108,31 +108,36 @@ export class CalculatedPropertyRule extends Rule {
...
-			var newList = newValue;
+			if (!this.property.isInited(obj)) {
+				Property$init(this.property, obj, newValue);
+			}
+			else {
+				// re-calculate the list values
+				var newList = newValue;
 
-			// compare the new list to the old one to see if changes were made
-			var curList = this.property.value(obj) as ObservableArray<any>;
```
tskimmett, line 115: This would cause infinite recursion because setting the calculated value for a list required obtaining the value, which would cause the rule to rerun, which would bring us back here.

src/calculated-property-rule.ts
```diff
@@ -108,31 +108,36 @@ export class CalculatedPropertyRule extends Rule {
 
 		// modify list properties to match the calculated value instead of overwriting the property
 		if (this.property.isList) {
-			// re-calculate the list values
-			var newList = newValue;
+			if (!this.property.isInited(obj)) {
+				Property$init(this.property, obj, newValue);
+			}
```
tskimmett, line 113: This clause prevents us from running the "list update" logic if the property is not yet initialized.

src/initilization-context.ts
```diff
@@ -14,23 +14,41 @@ export class InitializationContext {
...
 		});
 	}
 
-	ready(callback: () => void) {
-		if (this.tasks.size === 0)
+	get canProcessQueue() {
+		return this.tasks.size === 0;
+	}
+
+	private processWaitingQueue() {
```
tskimmett, line 41: Did a little refactoring of this class to make things easier to follow. Added `canProcessQueue` and `processWaitingQueue()`

src/initilization-context.ts
```diff
@@ -14,23 +14,41 @@ export class InitializationContext {
...
+		if (this.canProcessQueue) {
+			while (this.waiting.length > 0 && this.canProcessQueue) {
+				const done = this.waiting.shift();
+				done();
+			}
+		}
+	}
+
+	whenReady(callback: () => void) {
```
tskimmett, line 50: Renamed `ready` method`.

src/calculated-property-rule.ts
mattheyan: @tskimmett Do you know if the same behavior exists in the legacy codebase? If not, since this code is the same, I wonder what the difference is (maybe a missing short-circuit somewhere)?

src/calculated-property-rule.ts
mattheyan: Makes sense to me! Was this part of the fix, or just a related optimization?

src/initilization-context.ts
mattheyan: 👍

src/calculated-property-rule.ts
tskimmett: I'm not sure if it exists in legacy. I don't think we have any way to generate calculated list properties in the product though. I only discovered this issue when writing a test.

src/calculated-property-rule.ts
tskimmett: Not part of the fix, just a bug I discovered when writing tests.

src/initilization-context.ts
```diff
@@ -48,6 +66,6 @@ export class InitializationContext {
 	}
 
 	get readyPromise() {
```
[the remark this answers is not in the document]
mattheyan: The type in the name kind of bothered me, but its only used internally, so not worth worrying about.

src/calculated-property-rule.ts
mattheyan: 👍

src/calculated-property-rule.ts
mattheyan: Ah, good point. Only thing I could think of is lookups, but I'm guessing you're not using calculations for that, but manually managing lists?

tskimmett: Right, lookups can't be calculated (yet,, ;)

src/initilization-context.ts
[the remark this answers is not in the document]
tskimmett: Yeah, it was originally `readyPromise` so as not to conflict with the `ready` method. I can change it now though.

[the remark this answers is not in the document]
mattheyan: Cool. I think `context.ready.then(...)` reads pretty well.

src/initilization-context.ts
```diff
@@ -14,23 +14,41 @@ export class InitializationContext {
 		this.valueResolver = valueResolver;
 	}
 
+	/**
+	 * Prevents any waiting callbacks from being executed.
+	 * @returns A callback which removes your delay.
+	 */
+	delayQueue() {
```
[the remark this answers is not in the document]
tskimmett: **Slack conversation**
> taylor  11:55 AM
> I'm just getting around to looking at your feedback on the modeljs PR, re: `delayQueue`. I agree the pattern where you return a "reversal function" is a bit clunky. I'm not opposed to your "resolvable" object idea `let delay = context.delayQueue(); ... ; delay.resolve()`, but just wanted to float another idea:
> ```js
> const delayId = context.delayQueue();
> ...
> context.clearDelay(delayId);
> ```
> 
> bryan  11:55 AM
> Sort of like `setTimeout` and `setInterval`
> 
> taylor  11:55 AM
> Yeah
> 
> bryan  11:57 AM
> Interesting idea. Its kind of reminiscent of asynchronicity, but in this case its always used synchronously, right?
> 
> taylor  11:57 AM
> Yeah, it has to be synchronous
> 
> bryan  11:58 AM
> Is the usage pattern such that it could be something that took a callback, so it was entirely wrapped up in that call? Something like `context.something(() => { do stuff... });`?
> 11:59
> And something would internally set the "delay", execute the callback, then "release" it?
> 
> taylor  12:00 PM
> Hmm interesting idea, let me see
> 12:00
> I think that should work
> 12:01
> It could be an alternate signature for wait
> 
> bryan  12:01 PM
> Yeah, I was wondering the same thing
> 12:02
> The name seems kind of weird though, since its not async
> 
> taylor  12:03 PM
> `context.execute()`?
> 
> bryan  12:04 PM
> Yeah, that's what I was thinking... :thinking_face:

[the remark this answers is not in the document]
tskimmett: I've pushed the refactor of `delayQueue`.

src/entity.ts
```diff
@@ -57,11 +57,13 @@ export class Entity {
 			type.register(this);
 
 			// Initialize existing entity with provided property values
-			if (!isNew && properties)
-				this.init(properties, context);
+			if (!isNew && properties) {
+				// We need to pause processing of callbacks to prevent publishing entity events while still processing the state graph
+				context.execute(() => this.init(properties, context));
+			}
```
tskimmett, line 63: See use of new `execute` method.

src/type.ts
```diff
@@ -505,18 +505,24 @@ export function Type$createOrUpdate(type: Type, state: any, contextOrResolver: I
...
+		else {
+			// Cast the jstype to any so we can call the internal constructor signature that takes a context
+			// We don't want to put the context on the public constructor interface
+			const Ctor = type.jstype as any;
+			// Construct an instance using the known id if it is present
+			instance = (id ? new Ctor(id, state, context) : new Ctor(state, context)) as Entity;
+		}
+		return instance;
+	});
+
```
tskimmett, line 525: See use of new `execute` method.

src/initilization-context.ts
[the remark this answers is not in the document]
tskimmett: `whenReady` can potentially execute the callback immediately/synchronously, which is desirable in the event initialization has no async operations (calling an entity constructor). If everything used the promise, then it would default to async.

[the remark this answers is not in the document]
mattheyan: Makes sense 👍

src/initilization-context.ts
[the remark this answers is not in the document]
mattheyan: LGTM

mattheyan (approved):

PR #35, opened 2020-09-24
fix issues with nested formats

tskimmett:
This PR fixes two problems:
1. A reference property (type is an entity type) with a format does not format token values if the property type is listed after the containing type in model options (test: `entity.unit.ts`: `uses format of properties targeted by tokens`)
2. A validation rule whose message references a property with a dynamic label does not collect nested format paths as predicates (test: `validation-rule.unit.ts`: `nested format paths are added as predicates to rule referencing property with dynamic label`)


src/entity.unit.ts
```diff
@@ -337,5 +302,11 @@ describe("Entity", () => {
 			const movie = new Types.Movie(Alien);
 			expect(movie.toString("[Title] - [Director.FirstName] [Director.LastName]", v => `'${v}'`)).toBe("'Alien' - 'Ridley' 'Scott'");
 		});
+
+		it("uses format of properties targeted by tokens", () => {
```
tskimmett, line 306: Test of **problem 1**

src/model.ts
```diff
@@ -21,7 +21,8 @@ export class Model {
 
 	readonly entityRegistered: EventSubscriber<Model, EntityRegisteredEventArgs>;
 
-	private _ready: (() => void)[];
+	private _readyCallbacks: (() => void)[];
```
tskimmett, line 24: Rename the list of callbacks.

src/model.ts
```diff
@@ -277,30 +278,36 @@ export class Model {
...
 	}
 
-	ready(init: () => void): void {
-		if (this._ready) {
-			this._ready.push(init);
-		}
-		else {
-			init();
-		}
+	/**
```
tskimmett, line 301: Added a new parameter which allows a given callback to be invoked immediately if the queue of "ready callbacks" is already being processed.

src/validation-rule.unit.ts
```diff
@@ -61,6 +61,54 @@ describe("validation-rule", () => {
 		(instance as any).Number = 2;
 		expect(instance.meta.conditions[0].condition.message).toBe("Property Prop 2 is not valid");
 	});
+
```
tskimmett, line 64: Test of **problem 2**.

src/format.ts
```diff
@@ -231,13 +230,15 @@ export class ModelFormat<T extends Entity> extends Format<T> {
 									propertyDefaultFormat = lastProperty.format;
 								}
 
+								// Ensure the target format is compiled so we can collect its paths
```
tskimmett, line 233: Fixes **problem 2**.

src/property.ts
```diff
@@ -148,9 +148,7 @@ export class Property implements PropertyPath {
 				// Specifier
 				if (typeof (options.format) === "string") {
 					let format = options.format;
-					targetType.model.ready(() => {
-						this.format = targetType.model.getFormat(this.propertyType, format);
-					});
+					this.format = targetType.model.getFormat(this.propertyType, format);
```
tskimmett, line 151: Immediately create and assign the format to the property when extending the model. This means when the model is ready, all properties that should have formats defined will have a value for the `format` property.

src/format.ts
```diff
@@ -184,8 +183,8 @@ export class ModelFormat<T extends Entity> extends Format<T> {
 		this.type = type;
 		this.formatEval = formatEval;
 
-		// Compile the model format
-		this.compile();
+		// Compile the model format when the model is ready
```
tskimmett, line 186: Only compile the format when the model is ready to ensure properties referenced by tokens have been created/initialized. `enqueueWhileProcessing: false` allows the format to be compiled immediately if it is created after the model is ready, but the callback queue is currently being processed.

src/format.ts
```diff
@@ -7,7 +7,7 @@ import { evalPath, getConstructorName } from "./helpers";
...
...u1ff4\u1ff6-\u1ffc\u2071\u207f\u2090-\u2094\u2102\u2107\u210a-\u2113\u2115\u2119-\u211d\u2124\u2126\u2128\u212a-\u212d\u212f-\u2139\u213c-\u213f\u2145-\u2149\u214e\u2183-\u2184\u2c00-\u2c2e\u2c30-\u2c5e\u2c60-\u2c6f\u2c71-\u2c7d\u2c80-\u2ce4\u2d00-\u2d25\u2d30-\u2d65\u2d6f\u2d80-\u2d96\u2da0-\u2da6\u2da8-\u2dae\u2db0-\u2db6\u2db8-\u2dbe\u2dc0-\u2dc6\u2dc8-\u2dce\u2dd0-\u2dd6\u2dd8-\u2dde\u3005-\u3006\u3031-\u3035\u303b-\u303c\u3041-\u3096\u309d-\u309f\u30a1-\u30fa\u30fc-\u30ff\u3105-\u312d\u3131-\u318e\u31a0-\u31b7\u31f0-\u31ff\u3400-\u4db5\u4e00-\u9fc3\ua000-\ua48c\ua500-\ua60c\ua610-\ua62b\ua640-\ua65f\ua662-\ua66e\ua680-\ua697\ua722-\ua788\ua78b-\ua78c\ua7fb-\ua801\ua803-\ua805\ua807-\ua80a\ua80c-\ua822\ua840-\ua873\ua882-\ua8b3\ua8d0-\ua8d9\ua900-\ua925\ua930-\ua946\uaa00-\uaa28\uaa40-\uaa42\uaa44-\uaa4b\uaa50-\uaa59\uac00-\ud7a3\uf900-\ufa2d\ufa30-\ufa6a\ufa70-\ufad9\ufb00-\ufb06\ufb13-\ufb17\ufb1d\ufb1f-\ufb28\ufb2a-\ufb36\ufb38-\ufb3c\ufb3e\ufb40-\ufb41\ufb43-\ufb44\ufb46-\ufbb1\ufbd3-\ufd3d\ufd50-\ufd8f\ufd92-\ufdc7\ufdf0-\ufdfb\ufe70-\ufe74\ufe76-\ufefc\uff10-\uff19\uff21-\uff3a\uff41-\uff5a\uff66-\uffbe\uffc2-\uffc7\uffca-\uffcf\uffd2-\uffd7\uffda-\uffdc]*)(:(.+?))?\]/ig;
```
tskimmett, line 10: I refactored this to be a regex factory to avoid having to deal with resetting `lastIndex`. Not _entirely_ necessary, but the alternative is kinda nasty.

mattheyan: Yeah, that is annoying and can lead to bugs. I wish it was opt-in behavior.

src/format.ts
```diff
@@ -231,13 +230,15 @@ export class ModelFormat<T extends Entity> extends Format<T> {
 									propertyDefaultFormat = lastProperty.format;
 								}
 
+								// Ensure the target format is compiled so we can collect its paths
+								if (propertyDefaultFormat instanceof ModelFormat)
+									propertyDefaultFormat.compile();
+
 								// If the path references one or more entity properties, include paths for the property format. Otherwise, just add the path.
-								var lastIndex = formatTemplateParser.lastIndex;
```
mattheyan, line 235: What's this all about?

test/resource/en.js
```diff
@@ -0,0 +1,15 @@
+export default
```
mattheyan, line 1: There's already a `resource-en.ts`. Can that be used? Also, maybe we should consider shipping an english culture module by default as well?

tskimmett: Oh, I must have missed it.

src/format.ts
tskimmett: No idea honestly. The code between initializing `lastIndex` and restoring `formatTemplateParser.lastIndex` didn't use the regex, so it seemed like unnecessary code.

test/resource/en.js
tskimmett: Removed the resource file I added and used the existing `resource-en` module.

mattheyan (approved):

PR #37, opened 2020-09-30
Style/eslint cleanup

mattheyan:
- Added "no-trailing-whitespace" rule
- Enabled auto-fix on save
- Fixed eslint style violations (mostly auto-fix)


src/format.ts
```diff
@@ -354,10 +354,6 @@ export function createFormat<T>(model: Model, type: any, format: string): Format
 		var isPercentageFormat = format.match(/[%p]+/i);
 		var isIntegerFormat = format.match(/[dnfg]0/i);
 
-		var currencyDecimalDigits = model.$culture.numberFormat.CurrencyDecimalDigits;
-
-		var integerExpr = /^-?[0-9]{1,10}$/;
```
mattheyan, line 359: Unused variables...

src/entity.unit.ts
```diff
@@ -75,9 +73,8 @@ const Alien = {
 };
 
 describe("Entity", () => {
-	let model: Model;
```
mattheyan, line 78: Unused variable

src/calculated-property-rule.ts
```diff
@@ -1,6 +1,6 @@
 import { Rule, RuleInvocationOptions, RuleOptions } from "./rule";
 import { Type } from "./type";
-import { Property, PropertyRuleOptions, Property$init } from "./property";
```
mattheyan, line 3: Unused import

.eslintrc.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
 		"indent": ["error", "tab", { "SwitchCase": 1 }],
 		"object-curly-spacing": ["error", "always"],
 		"linebreak-style": ["error", "windows"],
+		"no-trailing-spaces": ["error"],
```
mattheyan, line 29: Added rule for trailing whitespace

.vscode/settings.json
```diff
@@ -1,4 +1,11 @@
 {
+    "editor.codeActionsOnSave": {
+        "source.fixAll": true
+    },
+    "eslint.validate": [
+        "javascript",
+        {"language":  "typescript",  "autoFix":  true  }
+    ],
```
mattheyan, line 8: Enable auto-fix on save, I *think* this config is ok... 🤷‍♂️

src/helpers.ts
```diff
@@ -321,6 +321,7 @@ export function flatMap<Input, Output>(input: Input[], callback: (i: Input) => O
  * @param input The input array
  */
 export function entries(obj: any): [string, any][] {
+	// eslint-disable-next-line one-var
```
mattheyan, line 324: Since this was a copy/pasted polyfill, I left it formatted as-is

src/globalization.ts
```diff
@@ -724,10 +724,10 @@ export function parseNumber(value: string, style: string = "Number" || "Integer"
 	if (style === "Currency") {
 		result = _parseNumber(value.replace(cultureInfo.numberFormat.CurrencySymbol, ""), cultureInfo) * sign;
 
-		// if there is a decimal place, check the precision isnt greater than allowed for currency. 
+		// if there is a decimal place, check the precision isnt greater than allowed for currency.
 		// Floating points in js can be skewed under certain circumstances, we are just checking the decimals instead of multiplying results.
 		var resultvalue = result.toString();
-		if (resultvalue.indexOf('.') > -1 && (resultvalue.length - (resultvalue.indexOf('.') + 1)) > cultureInfo.numberFormat.CurrencyDecimalDigits	) {
+		if (resultvalue.indexOf(".") > -1 && (resultvalue.length - (resultvalue.indexOf(".") + 1)) > cultureInfo.numberFormat.CurrencyDecimalDigits	) {
```
mattheyan, line 730: Apparently the current rule is to use double-quotes, which I realize is not consistent with the rules we use in other projects. I don't really care one way or another, but I'm hesitant to change it since it would show up as a violation all over the place, and to fix it we'd have to make lots of changes.

PR #38, opened 2020-09-30
Handle validation rule messages that are non-string

mattheyan:


tskimmett (changes requested):
These changes look good. Though I do think there is an opportunity for a couple low-effort tests of this behavior in `validation-rule.unit.ts`.

tskimmett (approved):
Looks great. Thanks for adding the test.

PR #41, opened 2020-11-13
return null when calling formatNumber with null

tskimmett:
Currently, `null` will be formatted as if it were `0`. There was a `isNaN()` check added to `formatNumber()` but `null` is considered to be a number. My guess is it was assumed `null` would be `NaN`.

### DIFF
--- a/src/globalization.ts
+++ b/src/globalization.ts
@@ -940,7 +940,7 @@ function expandNumber(number: number, precision: number, groupSizes: number[], s
  * @param cultureInfo The culture
  */
 export function formatNumber(number: number, format: string, cultureInfo: CultureInfo): string {
-	if (isNaN(number))
+	if (number === null || isNaN(number))
 		return null;
 
 	if (!format || (format.length === 0) || (format === "i")) {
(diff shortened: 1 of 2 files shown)


src/globalization.ts
```diff
@@ -940,7 +940,7 @@ function expandNumber(number: number, precision: number, groupSizes: number[], s
  * @param cultureInfo The culture
  */
 export function formatNumber(number: number, format: string, cultureInfo: CultureInfo): string {
-	if (isNaN(number))
+	if (number === null || isNaN(number))
```
mattheyan, line 943: I'm not 100% sure what the intention was with this check, but based on the following code it seems it assumes that `number` is a valid number after this check is performed. And, the bug is that it returns `0` unexpectedly, so that makes me think its intended to just catch any non-number, and if a non-number gets past the check the results will not be good (?). That being the case, I wonder if it would be useful to cover more bases by checking `typeof number != "number" || isNaN(number)`? The typeof check would catch null, as well as any other invalid value (string, object, undefined, etc.).

mattheyan (approved):

src/globalization.ts
tskimmett: Updated.

PR #42, opened 2020-12-02
Event scope revisited - bug fixes and module refactor

mattheyan:
**Problem:**
A lurking bug in the rules framework made it possible for a bad rules configuration to cause the rules engine to stop working, with no indication that a problem had occurred (i.e. no errors in the browser console).

**Solution:**
Fix the bug by making several changes to ensure that event scopes are always able to be disposed if an error occurs. When this happens some rules may have failed to run, leaving the model in a bad or unknown state, but at least rules can continue to run in the future and so the problems could be potentially be recovered from. Also, capture errors during rule execution and log a warning to the console to aid in troubleshooting. There are also changes to runaway rule detection, and other various changes including refactoring and testing changes, detailed below...

---

Rules can be configured in such a way that they never reach "steady state", essentially creating something akin to an infinite loop or infinite recursion. When this happens, the browser will eventually trigger an error at some point, and that behavior will vary from browser to browser, determined by factors such as call stack depth, number of statements, or even time elapsed. There was a lurking bug which resulted in the rules framework not being able to recover from an unexpected error if the script was terminated in a particular location (i.e. the EventScope constructor). All of this, combined with the error trapping behavior in `CalculatedPropertyRule` would result in a situation where rules would inexplicably stop working, which no other indication that a problem had occurred (i.e. no errors in the console).

**Summary of Changes:**
- **Moved the assignment of the "current" event scope into a try/catch**, to ensure that it is disposed of if an error occurs. This avoids the problem where the current event scope would never exit if an error was thrown from within the event scope constructor. This would have never been caught by a review, since the code in the constructor was pretty benign, so an error could only occur due to something unusual, like the browser terminating the script.
- **When disposing of event scopes after performing an action, always use the current scope**, to further ensure that all scopes are disposed. Logically, when an event scope's "perform" action is complete, any child scopes should have exited, so using the local event scope doesn't provide any logical benefit and only opens up the possibility of leaving the event scope in a corrupt state.
- **Wrap rule execution in a try/catch and log errors with `console.warn`**, including the rule name. This should help bring to light any errors that may have caused rule execution to halt unexpectedly, and hopefully point the user (or developer) in the direction of the rule that caused the problem.
- **Added event scope depth detection and optional limit**, since not all problems were covered by the "non-exiting event scopes" detection that was already in place. Depending on what rules are involved in a bad rules configuration, they may get "re-queued" repeatedly, or they may infinitely nest event scopes. The former was already detected, but the latter was not prior to these changes, and so rules would run indefinitely until whatever point that the browser chose to terminate the script.
- **Significant refactor of the event scope module, including converting to a class** (previously was a singleton pattern that used module state), as well as additional clean-up. The conversation of event scope to a class allows for it to be constructed per-model, which ensures model-level isolation. In practice since the browser is single threaded only one model can be running rules at a time (aside from the possibility of custom code explicitly running rules across models), but this makes the API consistent (pretty much everything else is model-scoped). Also, if an event scope were to be corrupted at least it wouldn't affect other models.
- **Added unit tests for the event scope module**. Previously there were not any explict tests of event scope. Now, there are some tests of basic use-cases, as well as the two known variations of "runaway rules" and the corresponding detection.
- **Added per-model configurability of "runaway rule" detection limits**. Previously these were hard-coded in the module. Now they can be specified when constructing a new model. See `ModelConfiguration`, which is the optional 2nd argument to the `Model` constructor.
- **Combined "abort" and "exit" events**, since they serve the same purpose and one of them will be called in all cases.
- **Implemented "preventDefault" for events**, and removed the unimplemented `stopPropagation` method. This was done so that the error event could be cancelled to avoid cluttering the console with error messages during unit test execution.

### DIFF
--- a/src/event-scope.ts
+++ b/src/event-scope.ts
@@ -45,23 +45,16 @@ export class EventScope {
 			// Invoke the callback
 			callback();
 			isDisposing = true;
-			scope.dispose({ abort: false });
+			this.current.dispose({ abort: false });
 		}
 		catch (e) {
-			if (!isDisposing) {
-				// Exit the event scope
-				scope.dispose({ abort: true });
-			}
+			if (!isDisposing)
+				this.current.dispose({ abort: true });
 		}
 		finally {
-			if (scope !== this.current) {
-				console.warn(`Exited non-current event scope ${scope._uid}.`);
-			}
-			else {
-				// Roll back to the closest active scope
-				while (this.current && !this.current.isActive) {
-					this.current = this.current.parent;
-				}
+			// Roll back to the closest active scope
+			while (this.current && !this.current.isActive) {
+				this.current = this.current.parent;
 			}
 		}
 	}
--- a/src/event-scope.ts
+++ b/src/event-scope.ts
@@ -1,19 +1,30 @@
 import { Event, EventSubscriber, EventSubscription } from "./events";
 import { getEventSubscriptions } from "./helpers";
 
-// Controls the maximum number of times that a child event scope can transfer events
-// to its parent while the parent scope is exiting. A large number indicates that
-// rules are not reaching steady-state. Technically something other than rules could
-// cause this scenario, but in practice they are the primary use-case for event scope.
-export const EventScope$nonExitingScopeNestingCount = 100;
+export interface EventScopeSettings {
+	/**
+	 * Controls the maximum number of times that a child event scope can transfer events to its parent
+	 * while the parent scope is exiting. A large number indicates that the consumer of the event scope
+	 * is likely cycling through the same activities repeatedly (i.e. infinite loop). This setting can
+	 * be used to cause the event scope to exit before the browser chooses to terminate the script in
+	 * a way that it may not be able to fully recover from.
+	 */
+	maxExitingTransferCount: number;
 
-// Controls the maximum depth that an event scope can reach. A large number indicates
-// that the consumer of the event scope is likely cycling through the same activities
-// repeatedly (i.e. infinite recursion) and the browser's "max stack" limit will
-// likely be reached at some point. This setting can be used to cause the event
-// scope to exit before the limit is reached, in order to avoid the browser
-// terminating the script in a way that it may not be able to fully recover from.
-export const EventScope$maxDepth = 1000;
+	/**
+	 * Controls the maximum depth that an event scope can reach. A large number indicates that the consumer
+	 * of the event scope is likely cycling through the same activities repeatedly (i.e. infinite recursion)
+	 * and the browser's "max stack" limit will likely be reached at some point. This setting can be
+	 * used to cause the event scope to exit before the limit is reached, in order to avoid the browser
+	 * terminating the script in a way that it may not be able to fully recover from.
+	 */
+	maxDepth: number;
+}
+
+export const EVENT_SCOPE_DEFAULT_SETTINGS: EventScopeSettings = {
+	maxExitingTransferCount: 100,
+	maxDepth: 1000
+};
 
 export interface EventScopeExitEventArgs {
 	abort: boolean;
@@ -27,23 +38,26 @@ export class EventScope {
 	current: EventScope = null;
 	isActive: boolean;
 
+	readonly settings: EventScopeSettings;
+
 	private readonly _uid: number;
 	private readonly _depth: number;
 
 	private readonly _onExit: EventSubscriber<EventScope, EventScopeExitEventArgs>;
 	private _exitEventVersion: number;
 
-	private constructor(parent: EventScope, isActive: boolean = false) {
+	private constructor(parent: EventScope, maxExitingTransferCount: number, maxDepth: number, isActive: boolean = false) {
 		this.parent = parent;
 		this.current = null;
 		this.isActive = isActive;
+		this.settings = { maxExitingTransferCount, maxDepth };
 		this._uid = ++__lastEventScopeId;
 		this._depth = parent === null ? 1 : parent._depth + 1;
 		this._onExit = new Event<EventScope, EventScopeExitEventArgs>();
 	}
 
-	static create(): EventScope {
-		return new EventScope(null, false);
+	static create({ maxExitingTransferCount = EVENT_SCOPE_DEFAULT_SETTINGS.maxExitingTransferCount, maxDepth = EVENT_SCOPE_DEFAULT_SETTINGS.maxDepth }: EventScopeSettings): EventScope {
+		return new EventScope(null, maxExitingTransferCount, maxDepth, false);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/event-scope.ts
```diff
@@ -1,177 +1,191 @@
...
 
-// TODO: Make `nonExitingScopeNestingCount` an editable configuration value
-// Controls the maximum number of times that a child event scope can transfer events
-// to its parent while the parent scope is exiting. A large number indicates that
-// rules are not reaching steady-state. Technically something other than rules could
-// cause this scenario, but in practice they are the primary use-case for event scope.
-const nonExitingScopeNestingCount = 100;
+export const EVENT_SCOPE_DEFAULT_SETTINGS: EventScopeSettings = {
+	maxExitingTransferCount: 100,
+	maxDepth: 1000
```
mattheyan, line 26: This could be set to a smaller, sane default. However, one potential down-side is that it seems possible for a somewhat large number of scopes to be created if there are a lot of rules that depend on one another, and a change triggers the whole "chain" to execute.

src/events.ts
```diff
@@ -1,10 +1,14 @@
 import { Functor$create, FunctorWith1Arg, FunctorItem } from "./functor";
 import { hasOwnProperty } from "./helpers";
 
-export class EventObject {
-	stopPropagation(): void {
-		// TODO: Implement 'stopPropagation()'?
-		throw new Error("Method 'stopPropagation' is not implemented.");
+export interface EventObject {
+	preventDefault(): void;
```
mattheyan, line 5: I wouldn't mind rolling this change back, since I only added it to support the unit tests. However, it seems simple enough. My only concern would be whether it creates some ambiguity or confusion for other events, since they wouldn't necessarily do anything in response to `preventDefault` being called.

src/event-scope.ts
```diff
@@ -1,177 +1,191 @@
...
+			throw new Error("Exceeded max scope event transfer.");
 		}
 
-		// Subscribe to the abort event
-		EventScope$current.onAbort.subscribe(callback.bind(thisPtr));
-	}
-}
+		// Move subscribers to the parent scope
+		subscriptions.forEach(sub => {
+			if (!sub.isOnce || !sub.isExecuted) {
```
tskimmett, line 178: Comment as to why `isOnce` subscriptions aren't transferred?

src/event-scope.ts
```diff
@@ -1,177 +1,191 @@
...
-	var scope = new EventScope();
-	try {
-		// Invoke the callback
-		if (thisPtr) {
-			callback.call(thisPtr);
-		}
-		else {
-			callback();
+		if (this._exitEventVersion !== undefined) {
+			this._exitEventVersion++;
```
tskimmett, line 184: Just want to make sure I'm understanding this correctly: `_exitEventVersion` only grows if we keep transferring events onto the same parent scope?

tskimmett (commented):
Just a few questions/minor comments.

src/event-scope.unit.ts
```diff
@@ -0,0 +1,203 @@
...
+	});
+	it("invokes the callback immediately if a scope was not already active", () => {
+		const scope = EventScope.create(EVENT_SCOPE_DEFAULT_SETTINGS);
+		let counter = 0;
+		scope.perform(() => {
+			expect(counter).toBe(0);
+		    counter++;
+		});
+		expect(counter).toBe(1);
+		counter++;
```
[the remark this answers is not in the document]
mattheyan: Nope. Probably a copy/paste bug. Good catch!

src/event-scope.unit.ts
```diff
@@ -0,0 +1,203 @@
...
+		let counter = 0;
+		scope.perform(() => {
+			expect(counter).toBe(0);
+			counter++;
+			scope.onExit(() => {
+				expect(counter).toBe(1);
+			});
+		});
+		expect(counter).toBe(1);
+		counter++;
```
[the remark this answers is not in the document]
mattheyan: Yes, I believe its there, along with some of the assertions in the `perform` callback, to verify that the `perform` callback is actually invoked prior to the code below it, and then `onExit` is invoked after `perform` exits, but before the code below it. But, I think that can probably be greatly simplified.

src/event-scope.ts
```diff
@@ -1,177 +1,191 @@
...
 export class EventScope {
-	parent: EventScope;
+	readonly parent: EventScope;
 
+	current: EventScope = null;
 	isActive: boolean;
 
-	private _exitEventVersion: number;
-	private _exitEventHandlerCount: number;
+	readonly settings: EventScopeSettings;
```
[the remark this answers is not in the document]
mattheyan: Good point. It is accessible via the model so they could be changed at any time, though in reality it shouldn't cause any problems. I could mark the interface members `readonly` and use `Object.freeze` in the constructor. 🤔

src/event-scope.ts
mattheyan: Ha, good question! I believe that's existing code/logic, so I'll dig into it to see if I can figure out, or remember why.

src/event-scope.ts
mattheyan: Hm... I'm honestly not sure about this logic. I'll take another look at it.

mattheyan: Yeah, that's the gist of it. I got hung up on the fact that the parent is exiting, but what happens is the list of exit event subscriptions keeps growing as the subscribers are invoked. For example, exiting a scope (ex: calculation of property A) triggers calculation of property B which exits and transfers its events to the parent, which includes a callback to trigger calculation of property C, which does the same thing for property B (or A), and then back to C again, etc...

src/event-scope.ts
mattheyan: Oh, right. "One time" subscriptions aren't transferred if they've already been run. This wasn't needed in Exo because the functor was added directly to the event subscription and it internally tracked this, whereas here we're directly subscribing the event handler. However, there is a logical error here in that if it is a one-time subscription that hasn't been run it will be subscribed as not a one-time subscription. But, it doesn't look like the one-time subscription concept is used (at least internally), so maybe I can just get rid of it. I may just tweak this logic though, since that's probably going to be simpler (and safer) than ripping out a bunch of code.

src/event-scope.unit.ts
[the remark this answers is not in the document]
tskimmett: Oh right, that makes sense.

src/event-scope.ts
tskimmett: Ok. That was kinda the scenario I was picturing in my head.

src/event-scope.ts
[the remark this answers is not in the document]
tskimmett: Not concerning enough to change imo. Gonna resolve this.

src/event-scope.unit.ts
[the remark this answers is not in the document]
mattheyan: I changed it to just increment in both callbacks and assert the counter was incremented twice, which I think accomplishes the same thing.

src/event-scope.ts
mattheyan: I went ahead and just simplified it since the "filter" and "once" options aren't being leveraged

tskimmett (approved):

PR #44, opened 2020-12-16
Added the ability to enforce AllowedValues when setting a value

Christopher-Masselli:
Added the enforcement of the AllowedValuesRule if preventInvalidValues is true when setting a value
Added a preventInvalidValues bool to AllowedValuesRule

### DIFF
--- a/src/property.ts
+++ b/src/property.ts
@@ -1022,8 +1022,8 @@ function Property$shouldSetValue(property: Property, obj: Entity, old: any, val:
 		throw new Error("Cannot set " + property.name + "=" + (val === undefined ? "<undefined>" : val) + " for instance " + obj.meta.type.fullName + "|" + obj.meta.id + ": a value of type " + (isEntityType(property.propertyType) ? property.propertyType.meta.fullName : parseFunctionName(property.propertyType)) + " was expected.");
 	}
 
-	for (let rule of property.rules) {
-		if (rule instanceof AllowedValuesRule && rule.preventInvalidValues && !rule.values(obj).includes(val) && !!val) {
+	for (const rule of property.rules) {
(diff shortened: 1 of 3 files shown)


src/property.ts
```diff
@@ -1020,6 +1022,12 @@ function Property$shouldSetValue(property: Property, obj: Entity, old: any, val:
 		throw new Error("Cannot set " + property.name + "=" + (val === undefined ? "<undefined>" : val) + " for instance " + obj.meta.type.fullName + "|" + obj.meta.id + ": a value of type " + (isEntityType(property.propertyType) ? property.propertyType.meta.fullName : parseFunctionName(property.propertyType)) + " was expected.");
 	}
 
+	for (let rule of property.rules) {
+		if (rule instanceof AllowedValuesRule && rule.preventInvalidValues && !rule.values(obj).includes(val) && !!val) {
```
tskimmett, line 1026: `!!val` would incorrectly skip validation of setting the property value to `false`. I think this is intended to be a `null/undefined` check?

src/property.ts
```diff
@@ -1020,6 +1022,12 @@ function Property$shouldSetValue(property: Property, obj: Entity, old: any, val:
 		throw new Error("Cannot set " + property.name + "=" + (val === undefined ? "<undefined>" : val) + " for instance " + obj.meta.type.fullName + "|" + obj.meta.id + ": a value of type " + (isEntityType(property.propertyType) ? property.propertyType.meta.fullName : parseFunctionName(property.propertyType)) + " was expected.");
 	}
 
+	for (let rule of property.rules) {
```
tskimmett, line 1025: Minor: you can use `const rule`.

tskimmett (changes requested):

src/property.ts
Christopher-Masselli: Yes, change made to explicitly check null/undefined

tskimmett (approved):

PR #45, opened 2021-01-25
Improve lookup form performance on page load

pkellz:
Description from **[Bug 18635](https://cognitoforms.visualstudio.com/Cognito%20Forms/_queries/edit/18635/?triage=true)**:
"A customer on Zendesk had noticed considerable delay before their public form can be interacted with on load. We have noticed anywhere from a ~4s-10s delay when reproducing. The form has a lookup field that references a form with > 7000 entries. This, in and of itself, shouldn't be a big issue. But the field also has 9 cascading filters. We believe this combination is the main contributing factor of the slowdown.

It appears the bulk of the time is spent looping through lookup entry index instances when initializing lookups. I think this can be optimized in the framework to eliminate this unnecessary work, which would reduce the delay considerable, if not altogether. This would have to be reproduced and the fix tested to prove that it would actually address the problem. Also, the same issue affects Vue forms, so it should at least be fixed there."

This PR reduces the time looping over indexes in `property-chain.js` by only attempting to loop over `rootType.known()` if the `priorProp` is not null.

After running several performance profiles both with the fix and without the fix, I found average page load times for both categories and it turns out that the changes improved performance by ~3 seconds (4300 entries x 9 cascading filters)

**_Page load times without changes_**
~12150 ms
~11970 ms
~13983 ms
~14115 ms
**Average without changes: ~13 seconds**

**_Page load times with changes_**
~13400 ms
~9200 ms
~8980 ms
~8720 ms
**Average with changes: ~10 seconds**

### DIFF
--- a/src/property-chain.ts
+++ b/src/property-chain.ts
@@ -88,16 +88,27 @@ export class PropertyChain implements PropertyPath {
 					let handler: PropertyChangeEventHandler;
 					let priorProp: Property = index > 0 ? props[index - 1] : null;
 					handler = args => {
-						this.rootType.known().forEach(known => {
-							if (this.testConnection(known, args.entity, priorProp)) {
-								(this.changed as EventPublisher<Entity, PropertyChangeEventArgs>).publish(known, {
-									entity: known,
-									property: args.property,
-									oldValue: args.oldValue,
-									newValue: args.newValue
-								});
-							}
-						});
+						if(priorProp) {
+							this.rootType.known().forEach(known => {
+								if (this.testConnection(known, args.entity, priorProp)) {
+									(this.changed as EventPublisher<Entity, PropertyChangeEventArgs>).publish(known, {
+										entity: known,
+										property: args.property,
+										oldValue: args.oldValue,
+										newValue: args.newValue
+									});
+								}
+							});
+						}
+						else 
+						{
+							(this.changed as EventPublisher<Entity, PropertyChangeEventArgs>).publish(args.entity, {
+								entity: args.entity,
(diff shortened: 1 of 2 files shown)


tskimmett (changes requested):
It would be nice to see some (unit) testing which emulates the scenario. It sounds like the test would assert whether we are making superfluous calls to `PropertyChain.testConnection()`. The most valuable part of the test would probably be seeing a simple reproduction of the model setup which exposes the issue.

My current understanding of this change is that it's a slight optimization for changes to the first property in a property chain, since we won't unnecessarily loop over pooled entities to see if they are linked to the changed entity via a `null` property, because `testConnection` would only return `true` if they are the same entity.

pkellz:
I'll look into creating the appropriate unit tests. 

And yes, you're understanding sounds correct. @mattheyan Can you confirm?

mattheyan:
Yes, it is an optimization, and that description is accurate. It was an oversight in the implementation that's lingered for years. I would say its a slight optimization in most cases (as evidenced by the fact that its lingered this long), but it depends on the scenario. In the case of a lookup with lots of entries, the impact is pretty significant (at least it definitely was in legacy). Patrick, do you remember if the same noticeable delay occurred in Vue forms as well. I seem to remember you confirmed the problem by measurement, but just curious if it was also noticeable to the user.

pkellz:
I believe I had noticed the Vue delay but it didn't feel outrageous. Overall, Vue was still noticeably quicker than legacy.

tskimmett (approved):

PR #46, opened 2021-02-15
Errors thrown in calculated rules are displayed

Christopher-Masselli:
Previously an error thrown in a calculated rule would not be printed to the console resulting in a generic error message. With this change we also print out the error that failed in the rule to provide more information.

before
`Error encountered while running rule Test.Calculated.`
after
`Error encountered while running rule Test.Calculated.`
`Error: Cannot set Test, "x" is not an allowed value.`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
+## [Unreleased]
+### Fixed
+- errors thrown during rule calculation are logged
+
 ## [0.8.8] - 2020-12-21
-### Added
-- ability to prevent property set if value would violate AllowedValuesRule
+### added
+- ability to prevent property set if value would violate allowedvaluesrule
 
 ## [0.8.7] - 2020-12-11
 ### Changed
--- a/src/calculated-property-rule.unit.ts
+++ b/src/calculated-property-rule.unit.ts
@@ -1,0 +1,59 @@
+import { Model } from "./model";
+
+import "./resource-en";
+
+function createModel(options) {
+	return new Promise((resolve) => {
+		let model = new Model(options);
+		model.ready(() => {
+			resolve(model);
+		});
+	});
+}
+describe("calculated-rule", () => {
+	test("Errors thrown in calculated rules are displayed to the user.", async () => {
+		let consoleOutputs = [];
+		window.console.warn = jest.fn((s)=> consoleOutputs.push(s.toString()));
+		const model = await createModel({
+			Test: {
+				Choice: {
+					label: "Choice",
+					default: {
+						dependsOn: "Text",
+						function: function() { return (this ? this.Text : null); }
+					},
+					allowedValues: {
+						ignoreValidation: true,
+						preventInvalidValues: true,
+						function: function() { return ["First", "Second", "Third"]; }
+					},
+					type: String
+				},
+				Choice_Choices: {
+					label: "Choice_Choices",
+					constant: [
+						{
+							Label: "First"
+						},
+						{
+							Label: "Second"
+						},
+						{
+							Label: "Third"
+						}
+					]
+				},
+				Text: {
+					label: "Text",
+					type: String
+				}
+			}
+		}) as any;
+		const Test = model.getJsType("Test");
+		var p = new Test({ Choice: "First", Text: "First" });
+
+		p.Text = "x";
+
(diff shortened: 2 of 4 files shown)


mattheyan:
When working on the `EventScope` fix I found that attempting to access the error details in the `catch` block could cause an error, which could corrupt the page "state". I was planning on adding the error details while I was in there, but I think I may have decided against it for that reason.

Christopher-Masselli:
@mattheyan Do you have a test setup or repo for an error being thrown in the catch?

mattheyan:
For the `EventScope` fix, the scenario was a large and/or complex form that caused the browser to terminate a running script because it reaches a "max execution" limit of some kind (details may vary from browser to browser). At the time I had a copy of a particular form that demonstrated the problem, but I'm not sure if I still have it somewhere or not. There are also tests (`event-scope.unit.ts`) to demonstrate the scenarios (the last two big ones toward the end of the file), and I _think_ I was also seeing different behavior in those tests when trying to log error details. I'm not 100% sure about that though, since the behavior in node.js and browsers is different (not surprisingly), but you could start with that at least.

Christopher-Masselli:
Do you have any recommendation to mitigate logging the error details producing an error, maybe lowering the "max execution" limit a little bit?

mattheyan:
Do you mean lowering the value of the framework setting in an attempt to avoid the browser terminating the script? Or something else?

I don't have any specific recommendations right now, since I don't recall 100% what the behavior was, just that it might cause unintended consequences. Have you tried running the unit tests I mentioned? We could also try to repro a scenario in the browser, but I would suggest checking the tests first since that should be easy. You should make sure they pass, but also check to see if we get different messages logged to the console with and without the changes. I seem to remember that it may have affected the error that would be caught and logged to the console. If that's the case, then I wouldn't worry about it since that's not a really big deal. But, if this change re-introduces the possibility of event scopes getting into a corrupt state again, that's a tougher call I think. If I had to make a bet I'd say it doesn't, but I'm just not sure. Getting the fix to work in all contexts was a bit fragile.

In terms of ideas... I don't think we can guarantee that the browser doesn't terminate a script (though we can reduce the probability by setting aggressively low thresholds), and I don't think we can detect the kind of error that would cause a problem (I believe its sort of an "observer effect" problem). Maybe one possibility would be to only log the error details when the current event scope's depth and version are under very small thresholds? I don't like the idea, but might be an option at least. @tskimmett Any thoughts or ideas?

tskimmett:
On the one hand I don't want error logging to cause this potentially corrupt state, but on the other hand, I don't like that the measures we've put in place effectively eliminate visibility into errors which occur during rule execution. Is it possible we could/should handle logging of the inner error at a lower level? It sounds like we can't really corrupt the event scope state by doing something _inside_ the scope (like logging), but I could be misunderstanding that. This is my mental picture for reference:

```
Event Scope
{
   run rule <-- throws error
   log error (possibly part of base Rule class?)
}

log 'Error encountered while running rule ...'
```

mattheyan:
I agree with that. I'm definitely a fan of making this change, it was something I personally made a note to try to get around to at some point, and planned on making the change as a part of that fix. I just want to make sure we don't unintentionally break something.

I don't _think_ the event scope fix made any changes to reduce logging (in fact, I believe some was added), and we're not doing anything with that to prevent logging from occurring per se. I just remember that when I was working on it, browsers behaved in very surprising ways when I tried to access (i.e. `console.warn`) error details when a script was being terminated. I _think_ what would happen is attempting to log the error details would cause another error to be thrown with an unhelpful error message (not sure exactly why). The impact of that might not be so terrible if that's it, but I have a nagging suspicion that I would have added the logging if that was the case. But, it could also be that I didn't add it because it made the logging in the test(s) less helpful.  I'll run the tests and see if I can gather anything from that.

mattheyan:
Logging the error details doesn't appear to have any impact on the tests. That's not conclusive evidence that there was another problem though.

Rather than go to the trouble of reproducing the original problem again (it was a huge pain after all), one option is we could just make the change realizing that it _could_ impact the event scope error recovery behavior. But, if it does hopefully our framework limits would be hit first anyway? 🤷

mattheyan:
I should clarify, its not just that its a huge pain, I'm not exactly sure how to reproduce at this point. I was able to do it at the time with a specific form. I probably have the info somewhere that I could track down, but it would take a little bit of digging.

mattheyan:
Unfortunately I was able to reproduce the problem... without the changes in this branch. 😧  I also created a repro org/form in prodtest so we'll always have it to refer back to. In the meantime, you may want to just move ahead with this change for now and treat this as a separate bug/regression.

tskimmett (approved):

tskimmett:
Thanks for checking up on everything Bryan. I know this issue is probably the bane of your existence at this point 😛 

@chris

PR #50, opened 2021-03-18
Serialize nulls

Nitrokitty:
Always seralize null values - not seralizing them will cause the server to add the property and consequentlly set the property's value to the property's default value

https://cognitoforms.visualstudio.com/Cognito%20Forms/_workitems/edit/19610

### DIFF
--- a/src/entity.unit.ts
+++ b/src/entity.unit.ts
@@ -69,11 +69,30 @@ function resetModel() {
 	});
 }
 
+// const Alien = {
+// 	Title: "Alien",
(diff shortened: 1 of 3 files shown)


src/entity.unit.ts
```diff
@@ -69,11 +69,30 @@ function resetModel() {
 	});
 }
 
+// const Alien = {
```
tskimmett: I pushed a commit to delete this.

tskimmett (approved):

PR #51, opened 2021-03-26
make autogeneration of property labels a configurable setting

Nitrokitty:
https://cognitoforms.visualstudio.com/Cognito%20Forms/_workitems/edit/19762/
https://cognitoforms.visualstudio.com/Cognito%20Forms/_workitems/edit/19876/


tskimmett:
What do you think about making this behavior configurable? It might be the best of both worlds, as I think autogenerating property labels makes for a more convenient API in many cases (one word property names), but obviously in our case is not preferable. We already have some configuration settings defined [here](https://github.com/cognitoforms/model.js/blob/fa31d1843a121a637f5af01f14f5042286dc4cbc/src/model.ts#L403).

tskimmett:
Several tests are failing as a result of the change. If we go for the configurable setting, we can default it to "autogenerate property labels" and the tests will work as they always have.

tskimmett (changes requested):
The test failures need to be addressed (see other comments for further discussion).

tskimmett (approved):

PR #52, opened 2021-03-29
Do not publish changes if there are no changes to publish

Nitrokitty:
For checkbox lookup, the field is publishing changes when the lookup list updates. However, there aren't any actual changes to the property value. Not publishing any changes, when there arent any changes to publish, fixes the issue.

https://cognitoforms.visualstudio.com/Cognito%20Forms/_workitems/edit/19874


tskimmett:
I've added onto this to ensure `ObservableArray` does not publish its internal `change` event when `raiseEvents()` is called and the list of queued changes is empty.

tskimmett (approved):

PR #55, opened 2021-04-12
join lists with ", " when formatting

Nitrokitty:
https://cognitoforms.visualstudio.com/Cognito%20Forms/_boards/board/t/Poseidon/Stories/?workitem=19899

### DIFF
--- a/src/format.ts
+++ b/src/format.ts
@@ -301,7 +301,7 @@ export class ModelFormat<T extends Entity> extends Format<T> {
 				if (this.formatEval)
 					value = this.formatEval(value);
 
-				if (typeof value !== "string" && value.length)
+				if (Array.isArray(value))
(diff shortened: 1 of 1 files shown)


src/format.ts
```diff
@@ -301,6 +301,9 @@ export class ModelFormat<T extends Entity> extends Format<T> {
 				if (this.formatEval)
 					value = this.formatEval(value);
 
+				if (typeof value !== "string" && value.length)
```
tskimmett, line 304:
```ts
if (Array.isArray(value))
```
would be a safer, more explicit check.

tskimmett (changes requested):

src/format.ts
Nitrokitty: forgot that was a thing haha. fixed!

PR #56, opened 2021-04-19
fix model initialization error due to timing of setting Type format

tskimmett:
`Type` was still setting its `format` within a model ready callback. `ModelFormat` had already been updated to defer format compilation until after the model is ready, so it was incorrect/unnecessary to _set_ the format inside a ready callback. See the added test for an example model which threw an error before the fix.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -291,7 +291,8 @@ export class Model {
 
 			// Complete pending model initialization steps
 			this._readyProcessing = true;
-			this._readyCallbacks.forEach(init => init());
+			for (const init of this._readyCallbacks)
+				init();
 			this._readyProcessing = false;
 			delete this._readyCallbacks;
 		}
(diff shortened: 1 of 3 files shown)


src/model.ts
```diff
@@ -291,7 +291,8 @@ export class Model {
 
 			// Complete pending model initialization steps
 			this._readyProcessing = true;
-			this._readyCallbacks.forEach(init => init());
+			for (const init of this._readyCallbacks)
```
tskimmett, line 294: This change was unrelated to the fix, but fixes a potential bug with ready callbacks. Array.forEach will not iterate over items added to the array during the iteration, which is a valid scenario when a ready callback queues another callback.

src/type.ts
```diff
@@ -67,9 +67,7 @@ export class Type {
 		// Set Format
 		if (format) {
 			if (typeof (format) === "string") {
-				this.model.ready(() => {
-					this.format = this.model.getFormat<Entity>(this.jstype, format);
-				});
+				this.format = this.model.getFormat<Entity>(this.jstype, format);
```
tskimmett, line 70: The fix.

Christopher-Masselli (approved):

PR #58, opened 2021-04-26
feat: add ability to "persist" an entity graph

tskimmett:
The only effect of calling this method would be marking entity meta objects as not "new". In essence, it is a way to reflect to the model that some outside notion of persistence was performed.

### DIFF
--- a/src/entity-serializer.ts
+++ b/src/entity-serializer.ts
@@ -2,7 +2,9 @@ import { isEntityType, Type } from "./type";
 import { Entity, EntityConstructorForType } from "./entity";
 import { Property } from "./property";
 import { ObjectLookup, flatMap } from "./helpers";
-import { InitializationContext, AsyncValueResolver } from "./initilization-context";
+import { InitializationContext } from "./initilization-context";
+
+export type AsyncValueResolver = (instance: Entity, property: Property, value: any) => Promise<any> | void;
 
 export interface PropertySerializationResult {
 	key: string;
--- a/src/entity-serializer.ts
+++ b/src/entity-serializer.ts
@@ -2,7 +2,7 @@ import { isEntityType, Type } from "./type";
 import { Entity, EntityConstructorForType } from "./entity";
 import { Property } from "./property";
 import { ObjectLookup, flatMap } from "./helpers";
-import { InitializationContext } from "./initilization-context";
+import { InitializationContext, AsyncValueResolver } from "./initilization-context";
 
 export interface PropertySerializationResult {
 	key: string;
@@ -70,6 +70,7 @@ export class EntitySerializer {
 	private _propertyConverters: PropertyConverter[] = [];
 	private _propertyInjectors = new Map<Type | string, PropertyInjector[]>();
 	private _propertyAliases = new Map<Type | string, ObjectLookup<string>>();
+	private _valueResolvers: AsyncValueResolver[] = [];
 	private static defaultPropertyConverter = new PropertyConverter();
 
 	/**
@@ -97,6 +98,10 @@ export class EntitySerializer {
 		let aliases = this._propertyAliases.get(type) || {};
 		aliases[alias] = propertyName;
 		this._propertyAliases.set(type, aliases);
+	}
+
+	registerValueResolver(resolver: AsyncValueResolver) {
+		this._valueResolvers.push(resolver);
 	}
 
 	/**
@@ -215,5 +220,13 @@ export class EntitySerializer {
 
 		propName = this.getPropertyAliases(context.meta.type)[propName];
 		return context.meta.type.getProperty(propName);
+	}
+
+	resolveValue(context: Entity, property: Property, value: any): Promise<any> | void {
+		for (const resolve of this._valueResolvers) {
+			const result = resolve(context, property, value);
+			if (result)
+				return result;
+		}
 	}
 }
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -289,6 +289,32 @@ export class Entity {
 	serialize(): object {
 		return this.serializer.serialize(this);
 	}
+
+	persist() {
+		const visited = new Set<Entity>();
+
+		function _persist(entity: Entity) {
+			if (visited.has(entity))
+				return;
+
+			visited.add(entity);
+
+			const identifierProp = entity.meta.type.identifier;
+			if (identifierProp && identifierProp.value(entity) === null)
+				return;
+
+			entity.meta.isNew = false;
+			for (const property of entity.meta.type.properties.filter(p => isEntityType(p.propertyType))) {
+				const value = property.value(entity);
+				if (Array.isArray(value))
+					(value as Entity[]).forEach(item => _persist(item));
+				else if (value)
+					_persist(value);
+			}
+		}
+
+		_persist(this);
+	}
 }
 
 export interface EntityConstructor {
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -196,11 +196,17 @@ export class Entity {
 			properties = property;
 
 		if (!this._context) {
+			const wasNew = this.meta.isNew;
 			const context = new InitializationContext(true);
-			context.execute(() => {
-				this.updateWithContext(context, properties);
-			});
-			return context.ready;
+			context.execute(() => this.updateWithContext(context, properties));
+
+			const markPersistedWhenIdAssigned = () => {
+				if (wasNew && !this.meta.isNew)
+					this.markPersisted();
+			};
+
+			context.whenReady(markPersistedWhenIdAssigned);
+			return context.ready.then(markPersistedWhenIdAssigned);
 		}
 
 		// Set the specified properties
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -1,6 +1,6 @@
 import { Event, EventObject, EventSubscriber } from "./events";
 import { Format } from "./format";
-import { Type, EntityType, isEntityType, getIdFromState, Type$createOrUpdate } from "./type";
+import { Type, EntityType, isEntityType, getIdFromState } from "./type";
 import { InitializationContext } from "./initilization-context";
 import { ObjectMeta } from "./object-meta";
 import { Property, Property$init, Property$setter } from "./property";
@@ -17,6 +17,7 @@ export class Entity {
 	readonly accessed: EventSubscriber<Entity, EntityAccessEventArgs>;
 	readonly changed: EventSubscriber<Entity, EntityChangeEventArgs>;
 	private _context: InitializationContext = null;
+	readonly initialized: Promise<void>;
 
 	constructor(); // Prototype assignment *** used internally
 	constructor(type: Type, id: string, properties?: ObjectLookup<any>, context?: InitializationContext); // Construct existing instance with state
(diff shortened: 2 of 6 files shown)


src/entity.unit.ts
```diff
@@ -344,4 +353,70 @@ describe("Entity", () => {
 			expect(movie.toString("[Director]")).toBe("Ridley Scott $100,000.00");
 		});
 	});
+
+	it("setting identifier property of entity marks meta.isNew = false", () => {
```
mattheyan, line 357: Love the tests! I could read through them and get a sense of what the method is supposed to do and some of the edge cases considered. 👍

src/entity.ts
```diff
@@ -289,6 +289,32 @@ export class Entity {
 	serialize(): object {
 		return this.serializer.serialize(this);
 	}
+
+	persist() {
```
mattheyan, line 293: I wonder about the method name. It kind of sounds like it would actively do some sort of persistence, although anyone thinking that should obviously pause to consider how exactly it would do that. 😛  Anyway, what do you think? `markPersisted` came to mind as a possibility. Seems like ideally it would be more specific to the `isNew` property, but I can't think of a good option.

src/entity.ts
```diff
@@ -289,6 +289,32 @@ export class Entity {
...
+
+			visited.add(entity);
+
+			const identifierProp = entity.meta.type.identifier;
+			if (identifierProp && identifierProp.value(entity) === null)
+				return;
+
+			entity.meta.isNew = false;
+			for (const property of entity.meta.type.properties.filter(p => isEntityType(p.propertyType))) {
+				const value = property.value(entity);
```
mattheyan, line 308: Should we filter out child entities that have an identifier property in this loop? If the idea is that when an entity is persisted its children that it "owns" are also persisted, then I would think a child with an identifier property (whether it has a value or not) wouldn't be included, as well as its children. Of course, I think the only time this would matter would be if an entity was persisted with a reference to a non-persisted entity. 🤷‍♂️

src/entity.ts
tskimmett: Method is now called `markPersisted`.

src/entity.ts
tskimmett: `markPersisted()` no longer traverses nested, identifiable entities.

src/entity-serializer.ts
```diff
@@ -99,6 +100,10 @@ export class EntitySerializer {
 		this._propertyAliases.set(type, aliases);
 	}
 
+	registerValueResolver(resolver: AsyncValueResolver) {
```
tskimmett, line 103: `AsyncValueResolvers` are now registered as a global concept for a given model. This avoids the need for users to pass resolver functions when creating entities.

src/entity.ts
```diff
@@ -63,17 +64,21 @@ export class Entity {
...
 
-				// Set values of new entity for provided properties
-				if (isNew && properties)
-					this.updateWithContext(context, properties);
+					// Set values of new entity for provided properties
+					if (isNew && properties)
+						this.updateWithContext(context, properties);
+
+					context.whenReady(resolve);
+				});
```
tskimmett, line 81: Because entity initialization and updates always have the possibility for async value resolution, the `initialized` promise provides a way to wait for the instance to be "fully baked" if that's something the user needs. As evidenced by most of the entity unit tests, if there is no asynchrony involved with the construction of a particular entity, a caller can assume the entity is baked immediately after constructing it.

src/entity.ts
```diff
@@ -152,9 +157,36 @@ export class Entity {
...
+				const Ctor = type.jstype as any;
+				// Construct an instance using the known id if it is present
+				instance = (id ? new Ctor(id, state, context) : new Ctor(state, context)) as Entity;
+			}
+			return instance;
+		});
+
+		return instance;
+	}
+
```
tskimmett, line 186: This function was pretty much a straight copy from `Type$createOrUpdate`.

src/entity.ts
```diff
@@ -152,9 +157,36 @@ export class Entity {
...
+	update(properties: ObjectLookup<any>): Promise<void>;
+	update(property: string, value: any): Promise<void>;
+	update(property: any, value?: any): Promise<void> {
 		let properties: ObjectLookup<any>;
```
tskimmett, line 190: `set()` renamed to `update()`.

src/entity.ts
```diff
@@ -163,6 +195,14 @@ export class Entity {
 		else
 			properties = property;
 
+		if (!this._context) {
+			const context = new InitializationContext(true);
+			context.execute(() => {
+				this.updateWithContext(context, properties);
+			});
+			return context.ready;
+		}
```
tskimmett, line 204: This facilitates updates being asynchronous by default.

src/entity.unit.ts
```diff
@@ -136,42 +151,109 @@ describe("Entity", () => {
...
+			model.serializer.registerValueResolver((entity, prop, value) => {
+				if (prop.name === "Budget" && value === "BUDGET_1")
+					return Promise.resolve(Budget1);
+			});
+			const movie = new Types.Movie({ ...Alien, Budget: "BUDGET_1" });
+			const expected = { ...Alien, Budget: Budget1 };
+			expect(movie.serialize()).not.toEqual(expected);
+			await movie.initialized;
+			expect(movie.serialize()).toEqual(expected);
+		});
```
tskimmett, line 166: New test.

src/entity.unit.ts
```diff
@@ -136,42 +151,109 @@ describe("Entity", () => {
...
+
+			expect(movie.Title).toBe("Alien 2");
+			expect(movie.Budget).toBeNull();
+
+			await updateTask;
+
+			expect(movie.Budget.serialize()).toEqual(Budget1);
+		});
+
+		it("should support circular async value resolution", async () => {
```
tskimmett, line 224: This test moved from `type.unit.ts`, since it was previously exercising the "update" aspect of `Type.create`, which is no longer a thing.

src/entity.unit.ts
```diff
@@ -136,42 +151,109 @@ describe("Entity", () => {
...
 			const movieBeforeSet = movie.Cast;
-			jest.spyOn(movie.serializer, "deserialize").mockImplementation(()=>undefined);
-			movie.set({ Cast: [{ FirstName: "Ridley", LastName: "Scott" }, { FirstName: "John", LastName: "Doe" }] });
+			jest.spyOn(movie.serializer, "deserialize").mockImplementation(() => undefined);
+			movie.update({ Cast: [{ FirstName: "Ridley", LastName: "Scott" }, { FirstName: "John", LastName: "Doe" }] });
 			expect(movie.Cast).toBe(movieBeforeSet);
 		});
+
+		it("allows waiting for asynchronous values to be set", async () => {
+			const Budget1 = { LineItems: [{ Label: "L1", Cost: 1000000 }] };
```
tskimmett, line 205: New test.

src/initilization-context.ts
```diff
@@ -1,17 +1,15 @@
 import { Entity } from "./entity";
 import { Property } from "./property";
 
-export type InitializationValueResolver = (instance: Entity, property: Property, value: any) => Promise<any>;
+export type AsyncValueResolver = (instance: Entity, property: Property, value: any) => Promise<any> | void;
```
tskimmett, line 4: Renamed this type to say what it means.

src/type.ts
```diff
@@ -78,38 +77,30 @@ export class Type {
...
+	get identifier(): Property {
 		if (this._identifier)
 			return this._identifier;
 		return this.baseType ? this.baseType.identifier : null;
 	}
 
-	set identifier(val) {
+	set identifier(val: Property) {
 		this._identifier = val;
 	}
```
tskimmett, line 89: Both `createSync` and `create` will throw an error if asked to "create" an entity which already exists. This means the type for which you want to create a new instance already has an object pooled under the identifier specified in the provided `state`.

src/type.unit.ts
```diff
@@ -17,109 +17,95 @@ describe("Type", () => {
...
+				}
+			});
 
-		// ensure entity instances already exist and are pooled
-		await model.types.Entity.create(entities.a, siblingResolver);
-		let root = await model.types.Entity.create(entities.a, siblingResolver);
-		expect((root as any).Sibling.Sibling).toBe(root);
+			await model.types.Entity.create({ Id: "x" });
+			expect(() => model.types.Entity.create({ Id: "x" })).toThrow(/already exists/);
+		});
```
tskimmett, line 109: ^^^ New test.

src/entity.unit.ts
```diff
@@ -344,4 +426,90 @@ describe("Entity", () => {
...
+
+				it("does not affect nested, sovereign entities", () => {
+					movie.markPersisted();
+					expect(movie.Director.meta.isNew).toBeTruthy();
+
+					movie.Director.update({
+						Id: "1",
+						// TODO: should Address be new after this update operation, even though Director's Id is getting assigned?
+						Address: { City: "Orlando", State: "Florida" }
+					});
```
tskimmett, line 502: See the TODO comment. Currently, in this example, `Address.meta.isNew === true` after this update occurs, because the id assignment technically occurs before the Address entity is created and attached to the Director. I don't know if that is a bad thing, but could definitely see it going either way.

src/entity-serializer.ts
```diff
@@ -216,4 +223,12 @@ export class EntitySerializer {
 		propName = this.getPropertyAliases(context.meta.type)[propName];
 		return context.meta.type.getProperty(propName);
 	}
+
+	resolveValue(context: Entity, property: Property, value: any): Promise<any> | void {
+		for (const resolve of this._valueResolvers) {
+			const result = resolve(context, property, value);
+			if (result)
+				return result;
```
tskimmett, line 231: As soon as one of the known resolvers returns something, use that as the value.

src/entity.ts
mattheyan: I had the thought that it might be nice to not expose it if a user would likely never have need to use the `initialized` promise, and handle it (somehow) via the `create` method. But, I don't think having the `initialized` property is a bad thing per se, and it probably isn't worth the effort (assuming it would be feasible).

src/entity.ts
mattheyan: Looks like the only difference was getting rid of the `contextOrResolver` logic and have it just be a context. Which I think was no longer needed since you made value resolvers a model concept.

src/entity.ts
```diff
@@ -163,6 +195,14 @@ export class Entity {
 		else
 			properties = property;
 
+		if (!this._context) {
+			const context = new InitializationContext(true);
```
mattheyan, line 199: Am I misreading this, or does this mean when you call `update` the context has "newDocument"=true? I don't know if it matters though...

src/entity.ts
mattheyan: Is it async by default or always async? This essentially re-enters the `updateWithContext` with a context if one isn't already present, which it looks to me like that would typically be the case when calling `update` directly, whereas when called via `updateWithContext` it would obviously already be defined and continue to the property setting below. Does this mean that `update` calls will always be async, or is there some condition where `_context` would _not_ be nulled out in `updateWithContext`?

src/entity.ts
tskimmett: You're reading it right. It pretty much just ensures any "owned" objects created via `update()` are considered to be "new".

```ts
entry.update({
  Section: {
    Field1: 'x',
    Field2: 'y'
  }
});

console.log(entry.Section.meta.isNew); // true
```

src/entity.ts
tskimmett: Yeah, that was poor wording on my part. The function always returns a `Promise`, so in that sense it is async always. I think what I was trying to get at is the fact that, in the event the user has not added any asynchrony to their model, you can treat `update()` as a synchronous operation.

```ts
entity.update({ Field1: 1, Field2: 2 });

console.log(entity.Field1, entity.Field2); // 1, 2
```

src/entity.ts
mattheyan: Ah, that makes sense

src/entity.ts
mattheyan: Good deal 👍

src/entity.unit.ts
mattheyan: Hmm, good question. I think from a user perspective I would expect it to no longer be new, since it appears as if setting the `Id` and setting the `Address` happen at the same time, or as a part of the same operation. Also, I could see this being the result of a server-side call which includes the objects persisted id as well as some additional data. Not sure how big of a deal it is though, especially depending on how much effort it would take to change. Also, is it true that you could always call `markPersisted` after setting the id?

src/type.ts
mattheyan: Makes sense to me

src/entity.unit.ts
tskimmett: I've modified the `update()` method to include a call to `markPersisted` after the update occurs so that owned entities created during an update that includes an id assignment are _not considered new_. Added a couple tests for this too.

mattheyan (approved):
LGTM!  😁

PR #60, opened 2021-05-12
Added afterPropertySet, listChanged global events and tests

Christopher-Masselli:
Added afterPropertySet - emits an event when a property that is not a list is changed
Added listChanged - emits an event when a property that is a list is changed

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -1,6 +1,6 @@
 import { Event, EventSubscriber } from "./events";
 import { replaceTokens, ObjectLookup } from "./helpers";
-import { EntityRegisteredEventArgs, Entity } from "./entity";
+import { EntityRegisteredEventArgs, Entity, EntityChangeEventArgs } from "./entity";
 import { Type, PropertyType, isEntityType, ValueType, TypeOptions, TypeExtensionOptions } from "./type";
 import { Format, createFormat } from "./format";
 import { EntitySerializer } from "./entity-serializer";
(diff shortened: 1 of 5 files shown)


src/model.ts
```diff
@@ -21,6 +21,8 @@ export class Model {
 	readonly $culture: CultureInfo;
 
 	readonly entityRegistered: EventSubscriber<Model, EntityRegisteredEventArgs>;
+	readonly afterPropertySet: Event<Entity, EntityChangeEventArgs>;
+	readonly listChanged: Event<Entity, EntityChangeEventArgs>;
```
tskimmett, line 25: These two properties should just be `EventSubscriber`, so external users don't get access to the `publish` method.

src/property.ts
```diff
@@ -1080,6 +1081,7 @@ function Property$setValue(property: Property, obj: Entity, currentValue: any, n
 			var eventArgs: PropertyChangeEventArgs = { entity: obj, property, newValue, oldValue };
 			(property.changed as EventPublisher<Entity, PropertyChangeEventArgs>).publish(obj, additionalArgs ? merge(eventArgs, additionalArgs) : eventArgs);
 			(obj.changed as Event<Entity, EntityChangeEventArgs>).publish(obj, { entity: obj, property, oldValue, newValue });
+			property.containingType.model.afterPropertySet.publish(obj, { entity: obj, property, newValue, oldValue });
```
tskimmett, line 1084: It may not really matter, but it makes more sense to me for the order of events to be the same in all cases. Here, the global event is published _after_ the object event, but for lists, the global event is published _before._

tskimmett (changes requested):
Changes look good, plus 👍👍  for the tests. Just a couple minor feedback points to address and we can get it merged.

src/property.ts
Christopher-Masselli: Moved to the top for both cases.

src/model.ts
Christopher-Masselli: Change made.

tskimmett (approved):

PR #63, opened 2021-06-03
feat: default value calculations run on existing entities

tskimmett:
When initializing an "existing" entity, default value calculation rules will run for properties with no value specified in the initial state for the entity.

```ts
// assuming Name is defaulted to "John Doe"
const e1 = await SomeType.create({ Id: "1" });
console.log(e1.Name);  // John Doe

const e2 = await SomeType.create({ Id: "2", Name: null });
console.log(e2.Name);  // null
```

### DIFF
--- a/src/calculated-property-rule.ts
+++ b/src/calculated-property-rule.ts
@@ -32,14 +32,13 @@ export class CalculatedPropertyRule extends Rule {
 			if (options.property) {
 				property = typeof options.property === "string" ? rootType.getProperty(options.property) as Property : options.property as Property;
 
-				if (!options.isDefaultValue) {
-					// indicate that the rule is responsible for returning the value of the calculated property
-					options.returns = [property];
-				}
-				else {
+				if (options.isDefaultValue) {
 					// Ensure the default value rule runs on init of a new instance
 					(options as RuleInvocationOptions).onInitNew = true;
 				}
+
+				// indicate that the rule is responsible for returning the value of the calculated property
+				options.returns = [property];
 			}
 
 			if (!name) {
@@ -68,8 +67,8 @@ export class CalculatedPropertyRule extends Rule {
 		// register the rule with the target property
 		this.property.rules.push(this);
 
-		// mark the property as calculated if the rule runs on property access
-		if (this.invocationTypes & RuleInvocationType.PropertyGet)
+		// mark the property as calculated if the rule runs on property access and is not a default value calculation
+		if (!options.isDefaultValue && this.invocationTypes & RuleInvocationType.PropertyGet)
 			this.property.isCalculated = true;
 	}
 
--- a/src/property.ts
+++ b/src/property.ts
@@ -840,73 +840,6 @@ export function Property$generateOwnProperty(property: Property, obj: Entity): v
 	});
 }
 
-// TODO: Get rid of `Property$_generateOwnPropertyWithClosure`...
-export function Property$generateOwnPropertyWithClosure(property: Property, obj: Entity): void {
-	let val: any = null;
-
-	let isInitialized = false;
-
-	var _ensureInited = function (): void {
-		if (!isInitialized) {
-			// Do not initialize calculated properties. Calculated properties should be initialized using a property get rule.
-			if (!property.isCalculated) {
-				Property$pendingInit(obj, property, false);
-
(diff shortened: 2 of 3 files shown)


src/property.ts
```diff
@@ -840,73 +840,6 @@ export function Property$generateOwnProperty(property: Property, obj: Entity): v
 	});
 }
 
-// TODO: Get rid of `Property$_generateOwnPropertyWithClosure`...
```
tskimmett, line 843: I did the TODO!

src/property.ts
```diff
@@ -984,16 +917,16 @@ export function Property$init(property: Property, obj: Entity, val: any): void {
...
-		Property$pendingInit(obj, property, true);
-
 		// Do not initialize calculated properties. Calculated properties should be initialized using a property get rule.
 		if (!property.isCalculated) {
 			Property$init(property, obj, Property$getInitialValue(property));
 		}
+
+		// Mark the property as pending initialization if it still has no underlying value to allow default calculation rules to run for it
+		if (obj.__fields__[property.name] === null)
+			Property$pendingInit(obj, property, true);
```
tskimmett, line 929: I adapted this from exoweb:
image

src/property.ts
```diff
@@ -984,15 +917,17 @@ export function Property$init(property: Property, obj: Entity, val: any): void {
...
-		// Mark the property as pending initialization
 		Property$pendingInit(obj, property, true);
 
 		// Do not initialize calculated properties. Calculated properties should be initialized using a property get rule.
 		if (!property.isCalculated) {
 			Property$init(property, obj, Property$getInitialValue(property));
+
+			// Mark the property as pending initialization if it still has no underlying value to allow default calculation rules to run for it
+			if (obj.__fields__[property.name] === null)
+				Property$pendingInit(obj, property, true);
```
tskimmett, line 930: I adapted this from exoweb:
image

src/property.ts
mattheyan: 🤣

src/property.ts
mattheyan: Oh, now I see that you changed the `isCalculated` logic to not set it when its a default value calculation 😁

src/calculated-property-rule.ts
```diff
@@ -32,14 +32,13 @@ export class CalculatedPropertyRule extends Rule {
...
-					options.returns = [property];
-				}
-				else {
+				if (options.isDefaultValue) {
 					// Ensure the default value rule runs on init of a new instance
 					(options as RuleInvocationOptions).onInitNew = true;
 				}
+
+				// indicate that the rule is responsible for returning the value of the calculated property
+				options.returns = [property];
```
mattheyan, line 41: The way I'm reading this, default value calculations would have never run for existing entities? Or, at least not unless one of their inputs changed. So, with this change, they will also run on first access if needed.

src/calculated-property-rule.ts
```diff
@@ -68,8 +67,8 @@ export class CalculatedPropertyRule extends Rule {
 		// register the rule with the target property
 		this.property.rules.push(this);
 
-		// mark the property as calculated if the rule runs on property access
-		if (this.invocationTypes & RuleInvocationType.PropertyGet)
+		// mark the property as calculated if the rule runs on property access and is not a default value calculation
+		if (!options.isDefaultValue && this.invocationTypes & RuleInvocationType.PropertyGet)
```
mattheyan, line 71: Why was this change needed? I'm guessing previously there was some reason why the default value calculation would not have run on first access. Was it just because the `returns` option wasn't set, or was there another reason as well?

mattheyan: Is it because the serialization/deserialization code ignores properties with `isCalculated` = true?

mattheyan (approved):
LGTM

src/calculated-property-rule.ts
tskimmett: Yes, if we didn't unset `isCalculated`, the default calculation rule would always overwrite the value. For example, when you load an existing entry that has a value other than what the default would be.

src/calculated-property-rule.ts
tskimmett: You're correct. Previously, default calculations would only run when a dependency changed for an existing entity.

PR #65, opened 2021-06-08
Fix format errors with tokenized labels to display the correct label

Christopher-Masselli:
use evaluateLabel() to properly get tokenized label


tskimmett (approved):
Looks like a pretty clear cut change. 💯

PR #67, opened 2021-07-28
Fix setting an empty list

Christopher-Masselli:
When updating a list with items to a list with no items, the list would not change since there wasn't anything implemented to handle setting an empty list.


tskimmett (changes requested):
Needs a test.

Christopher-Masselli:
Test added

tskimmett (approved):

PR #68, opened 2021-08-02
add: Type.createIfNotExists

tskimmett:
In order to support certain scenarios that previously worked, but now don't since `create` was changed to throw an error for known entity ids, a `createIfNotExists` method has been added to `Type`.


mattheyan (approved):
LGTM

PR #70, opened 2021-08-20
ensure async property values are initialized before publishing initExisting event

tskimmett:
When an entity is created, it should not be considered "fully initialized" until any asynchronously loaded property values have resolved, and are assigned. This timing issue was particularly problematic for rules which depended on a property whose value was asynchronously resolved. The order of events would look something like this:
1. entity begins creation (`Type.create()`)
2. a property with async value resolution enabled is encountered, we queue the task to load the data needed to init the property
3. the task completes, _but the property has not been initialized yet_
4. meanwhile, the rest of the entity is done initializing and the initialization context is deemed "ready"
4. `initExisting` event is published
5. "on init" rule runs, while value of our property is `null`
4. the async property is then initialized
5. the rule does not rerun because the property was not technically updated, but rather initialized
6. model is now in an inconsistent state

### DIFF
--- a/src/initilization-context.ts
+++ b/src/initilization-context.ts
@@ -30,9 +30,11 @@ export class InitializationContext {
 	wait(task: Promise<any>) {
 		this.tasks.add(task);
 		task.then(() => {
-			this.tasks.delete(task);
 			// process the queue asynchronously to allow additional tasks to be queued as a result of this one
-			Promise.resolve().then(() => this.processWaitingQueue());
+			Promise.resolve().then(() => {
+				this.tasks.delete(task);
+				this.processWaitingQueue();
+			});
 		});
 	}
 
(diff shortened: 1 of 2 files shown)


src/initilization-context.ts
```diff
@@ -30,9 +30,11 @@ export class InitializationContext {
 	wait(task: Promise<any>) {
 		this.tasks.add(task);
 		task.then(() => {
-			this.tasks.delete(task);
 			// process the queue asynchronously to allow additional tasks to be queued as a result of this one
```
tskimmett, line 33: As you can see in this comment, the original intention here was to ensure any tasks that depended on this task would be able to run _before_ we considered this task to be truly complete.

The specific action that raised the issue was [here](https://github.com/cognitoforms/model.js/blob/ef98f9ff3bcb83f249cce13b2e5a3cbb2fc56e42/src/entity.ts#L220).

The `then()` callback which deletes `task` from the task queue was guaranteed to run before any callbacks from the caller's frame, introducing the possibility that the "waiting queue" could be processed prematurely.

Christopher-Masselli (approved):

mattheyan (approved):
LGTM!

PR #71, opened 2021-08-30
Allow condition messages to be a function, allowing reactivity

Christopher-Masselli:
There wasn't a way to create a condition with a reactive message, so I allowed for a message to be passed in as a function. So messages where the property label has tokens are able to be made into functions making them reactive.

### DIFF
--- a/src/condition.ts
+++ b/src/condition.ts
@@ -10,7 +10,7 @@ import { Format } from "./format";
 
 export class Condition {
 	type: ConditionType;
-	message: string;
+	_message: string | Function;
 	targets: ObservableArray<ConditionTarget>;
 	source: Rule | Format<any>;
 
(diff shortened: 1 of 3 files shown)


src/condition.ts
```diff
@@ -10,7 +10,7 @@ import { Format } from "./format";
 
 export class Condition {
 	type: ConditionType;
-	message: string;
+	_message: string | Function;
```
tskimmett, line 13: Is it possible to make this a private field?

tskimmett, line 13: The function type can be more specific: `_message: string | () => string`

tskimmett (changes requested):

mattheyan:
What are the implications of this on the rules framework outside of a Vue.js context (where Vue reactivity wouldn't be a factor)?

src/condition.ts
Christopher-Masselli: Change made, the function had to be in parenthesis when used in a union type.

Christopher-Masselli:
> What are the implications of this on the rules framework outside of a Vue.js context (where Vue reactivity wouldn't be a factor)?

@mattheyan The format error will just behave the same way it did before.

tskimmett (approved):

PR #72, opened 2021-09-02
Add support for lists with formatting

Christopher-Masselli:
For lists `obj` in `convertToString()` was an array and we were not properly getting the format for the items in the list.

### DIFF
--- a/src/format.ts
+++ b/src/format.ts
@@ -278,35 +278,47 @@ export class ModelFormat<T extends Entity> extends Format<T> {
 		}
 
 		var result = "";
-		for (var index = 0; index < this.tokens.length; index++) {
-			var token = this.tokens[index];
-			if (token.prefix)
-				result = result + token.prefix;
-			if (token.path) {
-				var value = evalPath(obj, token.path);
-				if (value === undefined || value === null) {
-					value = "";
-				}
-				else if (token.format) {
-					let format: Format<any>;
-					if (token.format instanceof Format) {
-						format = token.format;
+		const convertTokens = (obj: T) => {
+			for (var index = 0; index < this.tokens.length; index++) {
+				var token = this.tokens[index];
+				if (token.prefix)
+					result = result + token.prefix;
+				if (token.path) {
+					var value = evalPath(obj, token.path);
+					if (value === undefined || value === null) {
+						value = "";
 					}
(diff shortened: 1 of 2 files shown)


src/format.unit.ts
```diff
@@ -62,4 +62,38 @@ describe("format", () => {
 		form.Text = "label2";
 		expect(error.message).toBe("label2 must be formatted as #,###.");
 	});
+
+	test("lists", async () => {
```
tskimmett, line 66: Can we improve the name of this test? If this fails, you would have to read through and understand the test code to figure out what the failure means.

`list items are formatted using the list property format`

src/format.unit.ts
```diff
@@ -62,4 +62,38 @@ describe("format", () => {
...
+					{
+						Text1: "Text1",
+						Text2: "Text2"
+					}]
+				}
+			},
+			"Form.Table": {
+				Text1: {
+					label: "Text1",
+					default: "Text1",
```
tskimmett, line 86: The defaults for `Text1` and `Text2` do not seem to be relevant for the test. The `Form.Table` type definition could likely be simplified to the following:
```ts
"Form.Table": {
    Text1: String,
    Text2: String
}
```

src/format.ts
```diff
@@ -278,35 +278,47 @@ export class ModelFormat<T extends Entity> extends Format<T> {
 		}
 
 		var result = "";
```
tskimmett, line 280: The pattern of calling `convertTokens` to accumulate the `result` value is a bit awkward now. It would simplify the logic if `convertTokens` were a pure function that returns a result instead of relying on the side effect of appending to `result`.

src/format.ts
```diff
@@ -278,35 +278,47 @@ export class ModelFormat<T extends Entity> extends Format<T> {
...
+					result = result + value;
+				}
 			}
-		}
+		};
+
+		if (Array.isArray(obj))
+			obj.forEach((item, index)=>{
+				convertTokens(item);
+				if (index !== obj.length - 1)
```
tskimmett, line 316: You can avoid the comma detection logic by using `map().join(", ")` (after the other feedback about `convertTokens` is addressed)

tskimmett (changes requested):

tskimmett (approved):

PR #73, opened 2021-09-14
Avoid prepending format token prefix for empty results

Christopher-Masselli:
**Problem**
We were adding the tokens prefix when ever it was available which was causing extra spaces when formatting if there were properties without a value.
`Ex. "[Prefix] [First] [MiddleInitial] [Last] [Suffix]" => " John  Doe "`
**Fix**
Avoid adding the prefix if the result and value are empty strings.

### DIFF
--- a/src/format.unit.ts
+++ b/src/format.unit.ts
@@ -88,4 +88,40 @@ describe("format", () => {
 		let form = await model.types.Form.create({}) as any;
 		expect(form.toString("[Table]")).toBe("Text2, Text2");
 	});
+
+	test("Prefixes are not prepended if there is no value and result", async () => {
+		var model = new Model({
+			"Form": {
+				Name: {
+					label: "Name",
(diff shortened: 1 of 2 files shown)


src/format.unit.ts
```diff
@@ -88,4 +88,40 @@ describe("format", () => {
 		let form = await model.types.Form.create({}) as any;
 		expect(form.toString("[Table]")).toBe("Text2, Text2");
 	});
+
+	test("Prefixes are not prepended if there is no value and result", async () => {
+		var model = new Model({
+			"Form": {
+				Name: {
+					label: "Name",
+					format: "[Prefix] [First] [MiddleInitial] [Last] [Suffix]",
```
tskimmett, line 97: What happens if the format looks like this?
```js
"'[First]' '[Last]'"
```

Christopher-Masselli: `'John'`
`Doe'`
`'`
Seems like an issue

PR #76, opened 2021-12-10
Ensure that the Boolean formatter is not affected by whitespace when parsing input

mattheyan:
- Trim both input and true/false values in the boolean formatter
- Add unit tests for Date/Number/Boolean custom format specifiers
- Add unit tests for low-level date and number formatting and parsing

### DIFF
--- a/src/calculated-property-rule.unit.ts
+++ b/src/calculated-property-rule.unit.ts
@@ -44,7 +44,7 @@ describe("CalculateRule", () => {
 		expect(consoleOutputs).toEqual(["Error encountered while running rule \"Test.Choice.Calculated\".", "Error: Cannot set Choice, \"x\" is not an allowed value."]);
 	});
 
-	describe("default value calculation", async () => {
+	describe("default value calculation", () => {
 		let model: Model;
 		let TestEntity: Type;
 		beforeEach(async () => {
--- a/src/format.ts
+++ b/src/format.ts
@@ -414,10 +414,11 @@ export function createFormat<T>(model: Model, type: any, format: string): Format
 				}
 			},
 			convertBack: function (str: string): boolean | FormatError {
-				if (str.toLowerCase() === trueFormat.toLowerCase()) {
+				str = str.trim();
+				if (str.toLowerCase() === trueFormat.trim().toLowerCase()) {
 					return true;
 				}
-				else if (str.toLowerCase() === falseFormat.toLowerCase()) {
+				else if (str.toLowerCase() === falseFormat.trim().toLowerCase()) {
 					return false;
 				}
 				else {
--- a/src/format.ts
+++ b/src/format.ts
@@ -402,6 +402,15 @@ export function createFormat<T>(model: Model, type: any, format: string): Format
 			nullFormat = formats.length > 2 ? formats[2] : "";
 		}
 
+		/**
+		 * Determines if the given string matches the given format option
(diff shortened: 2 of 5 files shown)


src/format.ts
```diff
@@ -402,6 +402,15 @@ export function createFormat<T>(model: Model, type: any, format: string): Format
...
 		}
 
+		/**
+		 * Determines if the given string matches the given format option
+		 * @param str The text to check
+		 * @param formatValue The true or false format option
+		 */
+		const isFormatMatch = (str, formatValue) => {
+			return str.trim().toLowerCase() === formatValue.trim().toLowerCase();
+		};
```
mattheyan, line 412: The main fix of the PR is to convert True/False values to lower-case before comparing to the input. I also added trimming of the input text, even though its being done in the `Format` class itself, since I'm not sure it necessarily should be done there. Finally, moved this logic into a reusable function.

src/format.unit.ts
```diff
@@ -88,4 +89,173 @@ describe("format", () => {
 		let form = await model.types.Form.create({}) as any;
 		expect(form.toString("[Table]")).toBe("Text2, Text2");
 	});
-});
\ No newline at end of file
+
```
mattheyan, line 92: Added tests for Date/Number/Boolean custom formatting. For each field type the tests check that null or whitespace is parsed as null, and formatting and parsing for each field sub-type (i.e.: Date/Time, Integer/Decimal/Percentage). Also, the `convertBack` calls include leading and trailing whitespace to ensure that the formatters aren't affected by that.

src/format.unit.ts
```diff
@@ -55,8 +56,8 @@ describe("format", () => {
 		});
 		let form = await model.types.Form.create({}) as any;
 		form.Text = "label1";
-		var val = model.types.Form.jstype.$Number.format.convertBack("test") as FormatError;
-		var error = val.createCondition(form, model.types.Form.jstype.$Number);
+		var val = model.types.Form.getProperty("Number").format.convertBack("test") as FormatError;
+		var error = val.createCondition(form, model.types.Form.getProperty("Number"));
```
mattheyan, line 60: Just changed this to call `.getProperty()` to avoid TypeScript compiler warnings.

src/calculated-property-rule.unit.ts
```diff
@@ -44,7 +44,7 @@ describe("CalculateRule", () => {
 		expect(consoleOutputs).toEqual(["Error encountered while running rule \"Test.Choice.Calculated\".", "Error: Cannot set Choice, \"x\" is not an allowed value."]);
 	});
 
-	describe("default value calculation", async () => {
```
mattheyan, line 47: This `async` keyword was unnecessary and jest printed a warning about it.

src/format.unit.ts
```diff
@@ -88,4 +89,173 @@ describe("format", () => {
...
+					PercentageField: {
+						label: "Percentage",
+						type: Number,
+						format: "P1"
+					}
+				}
+			});
+		});
+		test("parses null or whitespace as null", async () => {
+			let form = await model.types.Form.create({}) as any;
```
tskimmett, line 188: Not really a request to change it, but wanted to say the `form` initialization could be factored out since `beforeEach` supports `async`. I was not initially aware that was possible, so figured I'd share.

tskimmett (approved):

PR #81, opened 2022-03-17
Extending type creates new property if existing property types are different

Christopher-Masselli:
When extending a type by adding a property and the type already has an existing property we were just extending that existing property this causes issues if the new property has a different type than the existing property. To fix it if the types of the new property and existing property are different create a new property instead of extending the existing one.

### DIFF
--- a/src/type.ts
+++ b/src/type.ts
@@ -455,7 +455,7 @@ export class Type {
 
 					// Add Property
 					if (!property
-						|| (member.type && !isValueType(member.type) && !isValueType(property.propertyType) && property.propertyType.meta.fullName !== member.type)
+						|| (member.type && isEntityType(property.propertyType) && property.propertyType.meta.fullName !== member.type)
 						|| (member.type && isValueType(member.type) && isValueType(property.propertyType) && property.propertyType !== member.type)) {
 						// Type & IsList
 						let isList = false;
--- a/src/type.ts
+++ b/src/type.ts
@@ -454,7 +454,7 @@ export class Type {
 					let property = this.getProperty(name);
 
 					// Add Property
-					if (!property || property.containingType !== member.type) {
+					if (!property || (member.type && property.propertyType !== member.type)) {
 						// Type & IsList
 						let isList = false;
 						if (typeof (member.type) === "string") {
--- a/src/type.ts
+++ b/src/type.ts
@@ -454,7 +454,7 @@ export class Type {
 					let property = this.getProperty(name);
 
 					// Add Property
-					if (!property) {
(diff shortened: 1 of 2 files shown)


src/type.ts
```diff
@@ -454,7 +454,7 @@ export class Type {
 					let property = this.getProperty(name);
 
 					// Add Property
-					if (!property) {
+					if (!property || (member.type && property.propertyType !== member.type)) {
```
tskimmett, line 457: Isn't `member.type` technically a `string`, so that comparison would always come back as different?

src/type.ts
```diff
@@ -335,10 +335,10 @@ export class Type {
 	}
 
 	get properties(): Property[] {
-		let propertiesArray: Property[] = [];
-		for (var type: Type = this; type != null; type = type.baseType) {
+		let propertiesArray: Property[] = Object.values(this.__properties__);
+		for (var type: Type = this.baseType; type != null; type = type.baseType) {
 			for (var propertyName in type.__properties__) {
-				if (type.__properties__.hasOwnProperty(propertyName)) {
+				if (!this.__properties__.hasOwnProperty(propertyName)) {
```
tskimmett, line 341: It seems more correct to check a local map of the properties being gathered instead of `this.__properties__`. Right now, if there was a type hierarchy 3 levels deep, and the second level overrode a property from the top level, you would end up with both properties in the bottom type's list of `properties`.

tskimmett (changes requested):

src/type.ts
```diff
@@ -335,15 +335,15 @@ export class Type {
...
 	get properties(): Property[] {
-		let propertiesArray: Property[] = [];
-		for (var type: Type = this; type != null; type = type.baseType) {
+		let propertiesObject: { [name: string]: Property } = this.__properties__;
+		for (var type: Type = this.baseType; type != null; type = type.baseType) {
 			for (var propertyName in type.__properties__) {
-				if (type.__properties__.hasOwnProperty(propertyName)) {
-					propertiesArray.push(type.__properties__[propertyName]);
+				if (!propertiesObject.hasOwnProperty(propertyName)) {
+					propertiesObject[propertyName] = type.__properties__[propertyName];
```
tskimmett, line 342: We shouldn't be modifying the `__properties__` field.

tskimmett (changes requested):

src/type.ts
Christopher-Masselli: Updated, using spread to get `this.__properties__` in `propertiesObject`

tskimmett (approved):

PR #85, opened 2022-09-05
rely on PropertyGet event instead of InitNew for default calculations

tskimmett:
This fixes an issue where default property calculations referencing properties that are not listed as predicates (`dependsOn`) would run before said properties had been initialized. Since the defaulted property did not list the dependent property as a predicate, the default would not be recalculated when the other property was eventually set, after entity initialization.

This is not _strictly_ a bug, since property calculations shouldn't rely on properties without listing them as dependencies, but it does support a specific scenario in the product, and does not seem to have a downside.

### DIFF
--- a/src/property.ts
+++ b/src/property.ts
@@ -926,7 +926,7 @@ function Property$ensureInited(property: Property, obj: Entity): void {
 			Property$init(property, obj, Property$getInitialValue(property));
 
 			const underlyingValue = obj.__fields__[property.name];
-			// Mark the property as pending initialization if it still has no underlying value, or is the property's default, to allow default calculation rules to run for it
+			// Mark the property as pending initialization if it still has no underlying value, or is the property type's default, to allow default calculation rules to run for it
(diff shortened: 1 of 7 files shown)


src/property.ts
```diff
@@ -925,8 +925,9 @@ function Property$ensureInited(property: Property, obj: Entity): void {
 		if (!property.isCalculated) {
 			Property$init(property, obj, Property$getInitialValue(property));
 
-			// Mark the property as pending initialization if it still has no underlying value to allow default calculation rules to run for it
-			if (obj.__fields__[property.name] === null)
+			const underlyingValue = obj.__fields__[property.name];
+			// Mark the property as pending initialization if it still has no underlying value, or is the property type's default, to allow default calculation rules to run for it
+			if (underlyingValue === property.defaultValue || (property.isList && underlyingValue.length === 0))
```
mattheyan, line 930: Do you think it would be worth sanity checking for an array? List properties by default would be initialized as a new array, but maybe there is the possibility of someone setting their own default which could be null/undefined/junk?

mattheyan (approved):

PR #87, opened 2022-09-12
Fix for default value calculation based on a calculated property

mattheyan:
**Problem**:
In some cases, a default value rule may not run for an existing persisted entity when one of the rule's predicates calculates for the first time.

Calculations vs. Defaults:
Calculations generally run on first access in order to establish their initial value, since the property is not persisted and must obtain its value by running the calculation. However, default value rules are a little different - the property value is persisted, but the rule will run again if one of its predicates change. For a new entity, the default will always run on first access since there is no persisted value to use. Conversely for existing entities, if the entity has a persisted value for the property the rule will _**not**_ run unless one of its predicates fires a change event.

Calculations that are not accessed:
Typically a calculation will run for the first time as a result of first access of the property. For example, if its value is shown in the UI then the property must be accessed to get its value, which causes it to calculate its initial value. Because the calculation didn't previously have a value, it is treated as the initialization of the property and does not raise a change event. After the UI is initially rendered, if something causes the calculation to run again, it will then raise a change event, notifying subscribers that they should fetch the new value. However, this presumes that  something will access the calculation when the UI first renders. If the calculation is never accessed during initial render (ex: the field is hidden), then it won't establish its initial value, so when a predicate changes the result will be considered the calculation's initial value and will not raise a change event.

**Example**:

You have 3 fields on a page:
- `Name`: a text field
- `CleansedName`: a calculation which is not shown in the UI, which is based on `Name` (ex: converts the name to lower-case and removes whitespace and punctuation)
- `Username`: a text field which has a default of `= CleansedName ? CleansedName.Substring(0, 10) : "Username"`

If you load an existing record with a value of "John Doe" for `Name` and a `Username` of "johndoe", then change the name to "Test User", then the `Username` will still be "johndoe". This is because the `CleansedName` calculation is not accessed because it is not shown in the UI, so when `Name` changes it calculates for the first time and does not raise a change event. The expected behavior would be for the default to run and set `Username` to "testuser".

**Fix**:
In order to ensure that the default rule's calculated predicates fire a change event, we must ensure that the calculations are accessed when the object is initialized. This PR accomplishes that by running the default calculation and then discarding the value.

### DIFF
--- a/src/calculated-property-rule.ts
+++ b/src/calculated-property-rule.ts
@@ -89,10 +89,8 @@ export class CalculatedPropertyRule extends Rule {
 				// run the calculation and throw away the result.
 				const initialValue = args.entity.__fields__[this.property.name];
 				if (initialValue !== undefined) {
-					const calculateFn = this.ensureCalculateFn();
-
 					try {
-						calculateFn.call(args.entity);
+						this.calculateFn.call(args.entity);
 					}
 					catch (e) {
 					}
@@ -101,7 +99,7 @@ export class CalculatedPropertyRule extends Rule {
 		}
 	}
 
-	ensureCalculateFn() {
+	get calculateFn() {
 		let calculateFn: (this: Entity) => any;
 
 		// Convert string functions into compiled functions on first execution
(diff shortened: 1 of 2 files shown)


src/calculated-property-rule.ts
```diff
@@ -67,7 +71,29 @@ export class CalculatedPropertyRule extends Rule {
 			this.property.isCalculated = true;
 	}
 
-	execute(obj: Entity): void {
+	register() {
+		super.register();
+
+		// TODO: Restrict this to default values or apply to all calculation?
+		// Why is this here?
+		if (this.isDefaultValue) {
```
tskimmett, line 79: Removing this check definitely breaks tests.

tskimmett:
Description of issue:
- initialize default property to null
- update calculation's predicate
- calculation calculates, but does not trigger change since it only initialized
- access default property, still null because its calculation predicate never changed

src/calculated-property-rule.ts
mattheyan: Yeah, not sure why I was doubting the need for that. 🤷‍♂️

PR #91, opened 2022-10-19
fix: calculated list properties

tskimmett:
Prevent infinite recursion which occurs when calculating a list property for the first time. Since the property would not be initialized to a default (empty list), trying to get the current value of the property to perform the list update results in infinite recursion of the calculation rule.


src/calculated-property-rule.ts
```diff
@@ -102,6 +102,10 @@ export class CalculatedPropertyRule extends Rule {
 
 		// modify list properties to match the calculated value instead of overwriting the property
 		if (this.property.isList) {
+			// calculated lists need to be initialized to prevent infinite recursion when trying to get the current value below
+			if (Property$pendingInit(obj, this.property))
+				Property$init(this.property, obj, []);
```
mattheyan, line 107: Minor quibble. Is there a way to do this without using the "private" `Property$` functions? There is an `isInited` method on the property that you could use, though maybe the "pending init" concept is important here? And if the property isn't initialized then you could instead set the property value (the no change and update logic wouldn't be necessary). I'm not sure if the setter would do quite what you want in this case (looks like it might also require the property to be initialized), though if not then I think you could argue that's a bug.

src/calculated-property-rule.ts
```diff
@@ -1,6 +1,6 @@
 import { Rule, RuleInvocationOptions, RuleOptions } from "./rule";
```
mattheyan, line 1: Would you mind removing this unused import while you're in there? 😁

mattheyan (approved):
LGTM, just a couple of suggestions

src/calculated-property-rule.ts
tskimmett: I was able to use an alternate approach to fixing the issue, which was simply to allow the existing `Property$ensureInited` logic to run for a calculated list property, since it is invalid for a list property to not have an actual array value.

src/property.ts
```diff
@@ -921,8 +921,8 @@ function Property$ensureInited(property: Property, obj: Entity): void {
 	if (!obj.__fields__.hasOwnProperty(property.name)) {
 		Property$pendingInit(obj, property, true);
 
-		// Do not initialize calculated properties. Calculated properties should be initialized using a property get rule.
-		if (!property.isCalculated) {
+		// Do not initialize calculated properties. Calculated properties should be initialized using a property get rule. Lists must always be initialized however.
+		if (!property.isCalculated || property.isList) {
```
mattheyan, line 925: Question: Would this cause a calculation or default to start raising a change event where it wouldn't have previously?

tskimmett: Default calculations would not have `isCalculated = true`. For actually calculated lists, those were not used/working at all, so I don't think there is any room for regression there.

mattheyan: Gotcha. Yeah, that's good that there won't be a regression. But, is the behavior consistent between list and non-list calculated properties? Maybe this would cause a default to recalculate when a calculated list value was established for the first time? Just trying to think if there are any potential gotchas.

tskimmett: Yeah it would trigger a change event on the list the first time it calculates.

tskimmett: Had to revert back to the original solution which initializes the list from the calc rule in order to fix the initial calculation behavior.

PR #95, opened 2022-11-08
Only run default rule for list properties onInitNew

Christopher-Masselli:
Previously, list properties on existing entities could be calculated to their default value if the property was not specified when constructing the entity. This change effectively prevents a list property's default rule from running on property access, since the underlying value of a list property will never equal the default value. Instead, the list will be defaulted in response to the `initNew` event of the parent entity.

`console.log([] === []); // false`

### DIFF
--- a/src/entity.unit.ts
+++ b/src/entity.unit.ts
@@ -541,7 +541,7 @@ describe("Entity", () => {
 			expect(instance.Skills.length).toBe(2);
 			instance = await model.types.Person.create({ Skills: [] }) as any;
 			expect(instance.Skills.length).toBe(2);
-			instance = await model.types.Person.create({ Id: "test", Skills: [] }) as any;
+			instance = await model.types.Person.create({ Id: "test" }) as any;
 			expect(instance.Skills.length).toBe(0);
 		});
 	});
--- a/src/entity.unit.ts
+++ b/src/entity.unit.ts
@@ -56,7 +56,7 @@ function resetModel() {
 				required() {
 					return !isNaN(this.ReleaseYear);
 				},
-				dependsOn: "ReleaseDate",
+				dependsOn: "ReleaseDate"
 			},
 			Genres: "String[]",
 			Credits: {
@@ -369,7 +369,7 @@ describe("Entity", () => {
 		});
 
 		it("default values are run when serializing", async () => {
-			const defaultModel  = new Model({
+			const defaultModel = new Model({
 				Test: {
 					A: {
 						type: String,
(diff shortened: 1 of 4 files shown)


tskimmett:
I would update the title and description along these lines:

**only run default rule for list properties onInitNew**

> Previously, list properties on existing entities could be calculated to their default value if the property was not specified when constructing the entity. This change effectively prevents a list property's default rule from running on property access, since the underlying value of a list property will never equal the default value.
> 
> ```js
> console.log([] === []); // false
> ```

tskimmett:
Can you add a test which confirms the list is not defaulted when constructing an _existing_ entity without specifying the list in the initial state? [This seems like a good spot](https://github.com/cognitoforms/model.js/blob/master/src/entity.unit.ts#L457).

tskimmett (approved):

src/entity.unit.ts
```diff
@@ -533,6 +534,16 @@ describe("Entity", () => {
...
 			expect(instance.Skills.length).toBe(1);
 		});
+
+		it("only runs default rule for list properties onInitNew", async () => {
+			const model = new Model(PersonWithSkillsModel);
+			let instance = await model.types.Person.create({}) as any;
+			expect(instance.Skills.length).toBe(2);
+			instance = await model.types.Person.create({ Skills: [] }) as any;
+			expect(instance.Skills.length).toBe(2);
+			instance = await model.types.Person.create({ Id: "test", Skills: [] }) as any;
```
tskimmett, line 544: This needs to not have `Skills` in the initial state.

Christopher-Masselli: Removed

src/property.ts
```diff
@@ -927,7 +927,7 @@ function Property$ensureInited(property: Property, obj: Entity): void {
 
 			const underlyingValue = obj.__fields__[property.name];
 			// Mark the property as pending initialization if it still has no underlying value, or is the property type's default, to allow default calculation rules to run for it
-			if (underlyingValue === property.defaultValue || (property.isList && Array.isArray(underlyingValue) && underlyingValue.length === 0))
+			if (underlyingValue === property.defaultValue)
```
tskimmett, line 930: Discussed this with Bryan, and determined this logic is a bit confusing, since it is avoiding defaulting lists in a very subtle way. Suggestion is to add an explicit check for array, and comment:

```ts
// List properties are defaulted onInitNew instead of on access
if (underlyingValue === property.defaultValue && !Array.isArray(underlyingValue))
```

@mattheyan Feel free to comment with any additional feedback.

Christopher-Masselli: Updated

tskimmett (approved):

PR #96, opened 2022-11-11
fix: raise property change when default calculation rule runs for list

tskimmett:
With the changes to support calculated list properties, the behavior of _defaulted_ list properties changed such that the initial calculation of the list's default value did not trigger change for the list property itself. This behavior is inconsistent with default calculations for non-list properties, which do trigger a change event when they are defaulted for the first time.


Christopher-Masselli (approved):

mattheyan (approved):
LGTM!

PR #104, opened 2022-11-28
feat: 'init' model option

tskimmett:
Introduces a new model extension option called `init`, which is used to control the initialization behavior of a property. The function provided for the `init` option will be used to obtain the value with which to initialize the associated property. The initializer function is called only once, and is not reactive as a rule would be.

### DIFF
--- a/src/calculated-property-rule.ts
+++ b/src/calculated-property-rule.ts
@@ -1,6 +1,6 @@
 import { Rule, RuleInvocationOptions, RuleOptions } from "./rule";
 import { Type } from "./type";
-import { Property, Property$init, Property$pendingInit, PropertyRuleOptions } from "./property";
+import { Property, Property$init, PropertyRuleOptions } from "./property";
 import { Entity } from "./entity";
 import { ObservableArray, updateArray } from "./observable-array";
 import { RuleInvocationType } from "./rule-invocation-type";
@@ -112,15 +112,11 @@ export class CalculatedPropertyRule extends Rule {
 
 			// ensure the initial calculation of the list does not raise change events
 			// defaulting a list property should raise change events
-			if (!this.isDefaultValue && Property$pendingInit(obj, this.property))
+			if (!this.isDefaultValue && !this.property.isInited(obj))
 				Property$init(this.property, obj, newList);
 			else {
 				// compare the new list to the old one to see if changes were made
 				const curList = this.property.value(obj) as ObservableArray<any>;
-
-				// Prevent overwriting the current value with the default value
-				if (this.isDefaultValue && curList.length > 0)
-					return;
 
 				if (newList.length === curList.length) {
 					var noChanges = true;
--- a/src/property.ts
+++ b/src/property.ts
@@ -217,6 +217,37 @@ export class Property implements PropertyPath {
 				this.changed.subscribe(function(e) { options.set.call(this, e.newValue); });
 			}
 
+			// Init
+			if (options.init !== undefined) {
+				let initFn: (this: Entity) => any;
+				if (isPropertyValueFunction<any>(options.init)) {
+					initFn = options.init;
+				}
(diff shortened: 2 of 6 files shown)


src/calculated-property-rule.ts
```diff
@@ -112,16 +112,12 @@ export class CalculatedPropertyRule extends Rule {
 
 			// ensure the initial calculation of the list does not raise change events
 			// defaulting a list property should raise change events
-			if (!this.isDefaultValue && Property$pendingInit(obj, this.property))
+			if (!this.isDefaultValue && !this.property.isInited(obj))
 				Property$init(this.property, obj, newList);
 			else {
 				// compare the new list to the old one to see if changes were made
 				const curList = this.property.value(obj) as ObservableArray<any>;
```
tskimmett, line 120: Removing workaround logic for defaulting of lists

src/property.ts
```diff
@@ -625,6 +647,8 @@ export interface PropertyOptions {
 	/** An optional constant default value, or a function or dependency function object that calculates the default value of this property. */
 	default?: PropertyValueFunction<any> | PropertyValueFunctionAndOptions<any> | Value | Value[];
 
+	init?: PropertyValueFunction<any>;
```
tskimmett, line 650: Allowing functions only for simplicity.

src/property.ts
```diff
@@ -885,12 +909,14 @@ function Property$subArrayEvents(obj: Entity, property: Property, array: Observa
...
-function Property$getInitialValue(property: Property): any {
+function Property$getInitialValue(property: Property, obj: Entity): any {
 	// Constant
 	if (property.isConstant)
 		return typeof property.constant === "function" ? (property.constant = property.constant()) : property.constant;
 
-	var val = property.defaultValue;
+	var val = property.initializer
+		? property.initializer.call(obj)
+		: property.defaultValue;
```
tskimmett, line 919: Highlights the overlap between `default` (as it currently works) and `init`. Idea would be to diverge these such that `default` is used exclusively for _reactively_ updating a property.

src/calculated-property-rule.ts
```diff
@@ -142,7 +142,7 @@ export class CalculatedPropertyRule extends Rule {
 
 			// ensure the initial calculation of the list does not raise change events
 			// defaulting a list property should raise change events
-			if (!this.isDefaultValue && Property$pendingInit(obj, this.property))
+			if (!this.isDefaultValue && !this.property.isInited(obj))
```
tskimmett, line 145: Inconsequential change, just switched from `$pendingInit`.

src/property.ts
```diff
@@ -214,7 +215,28 @@ export class Property implements PropertyPath {
 
 			// Set
 			if (typeof options.set === "function") {
+				const property = this;
 				this.changed.subscribe(function(e) { options.set.call(this, e.newValue); });
```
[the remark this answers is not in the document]
tskimmett: This was prompted by an issue where back reference properties on section items were not set correctly, since the new `init` behavior for sections does not technically "change" the property.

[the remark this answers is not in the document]
tskimmett: Combined the property `changed` subscription with the rule.

mattheyan (approved):

src/property.unit.ts
```diff
@@ -24,4 +24,99 @@ describe("Property", () => {
...
+						init() {
+							return [{ Name: "Skill 1" }, { Name: "Skill 2" }];
+						}
+					}
+				},
+				Skill: { Name: String }
+			});
+			const instance = await model.types.Person.create({ Skills: [] }) as any;
+			expect(instance.serialize().Skills).toMatchObject([]);
+		});
```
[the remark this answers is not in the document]
tskimmett: Agreed. Added.

PR #107, opened 2022-12-07
fix: condition not applied to correct targets involving nested properties

andrewthebare:
Solution for [this bug](https://dev.azure.com/cognitoforms/Cognito%20Forms/_workitems/edit/27500) 
- replace the filter parameter with a check for equality with the last property inside the callback

This issue was that the filter was causing the `each` method to miss nested properties (like for when a calculation should show an error both on itself and on the field featured in the calculation). 

Removing the filter guarantees the callback to create an error is always called with every property, and the added if statement checks to see if the property is the lastProperty so that the error is only created for matching properties

### DIFF
--- a/src/condition-rule.unit.ts
+++ b/src/condition-rule.unit.ts
@@ -36,4 +36,64 @@ describe("ConditionRule", () => {
 		expect(p.meta.conditions).toHaveLength(1);
 		expect(p.meta.conditions[0].condition.message).toBe("Test error message.");
 	});
+
+	test("Nested Property", async () =>{
+		const model = await createModel({
+			Test: {
+				Section: {
+					type: "Section"
+				},
+				Text: {
+					type: String,
+					get: {
+						dependsOn: "Section.Text",
+						function() { return this.Section.Text; }
+					},
+					error: {
+						dependsOn: "Calculation",
+						function: function() {
+							if (this.Calculation !== null)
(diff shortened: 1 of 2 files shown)


tskimmett (changes requested):
I would appreciate a more informative description for this PR, answering questions along the lines of:

- What is the problem with the framework?
- How exactly is the behavior different from before?

src/condition-rule.unit.ts
```diff
@@ -36,4 +36,64 @@ describe("ConditionRule", () => {
...
+		var p = new Test({
+			Section: new Section({ Text: null }),
+			Calculation: new Calc("Test")
+		});
+		expect(p.meta.conditions).toHaveLength(0);
+		p.Section.Text = "x";
+		expect(p.meta.conditions).toHaveLength(2);
+		expect(p.meta.conditions[0].condition.message).toBe("Calculation Error Message");
+		expect(p.meta.conditions[1].condition.message).toBe("Test error message.");
+	});
```
tskimmett, line 98: Knowing how this test fails without the change would also help understand what's going on.

andrewthebare: I added this test to try to simulate the issue causing the [bug found here](https://dev.azure.com/cognitoforms/Cognito%20Forms/_workitems/edit/27500)

tskimmett (approved):

PR #109, opened 2022-12-27
Change property initialization to use update after establishing a default property value

mattheyan:
If properties are accessed and rules run when a property initializer is called (i.e. `init` model option), this can result in a few related problems.
- When there is a 2-way relationship between two types via properties on each type (ex: a child object that has a back-reference to its parent), then creating the child type may trigger an infinite cycle of object creation.
- Property access during initialization can cause an excessive amount of objects to be created, which can negatively impact performance or cause the browser to terminate the script.
- Property access during initialization can cause an object to be in an inconsistent state when it is ultimately introduced into the object graph, and since this is done via property "initialization", no change events will fire that could in theory get the model into a consistent state.

To avoid these problems, the initializer will instead be passed to `update` after the property's default value has been established.

### DIFF
--- a/src/property.ts
+++ b/src/property.ts
@@ -233,10 +233,7 @@ export class Property implements PropertyPath {
 				else
 					throw new Error(`Invalid property 'init' option of type '${getTypeName(options.init)}'.`);
 
-				const property = this;
-				this.initializer = function () {
-					return targetType.model.serializer.deserialize(this, initFn.call(this), property, new InitializationContext(true));
-				};
+				this.initializer = initFn;
 			}
 
 			// Default
@@ -913,26 +910,6 @@ function Property$subArrayEvents(obj: Entity, property: Property, array: Observa
 	});
 }
 
-function Property$getInitialValue(property: Property, obj: Entity): any {
-	// Constant
-	if (property.isConstant)
-		return typeof property.constant === "function" ? (property.constant = property.constant()) : property.constant;
-
-	var val = property.initializer
-		? property.initializer.call(obj)
-		: property.defaultValue;
-
-	if (Array.isArray(val)) {
-		val = ObservableArray.ensureObservable(val as any[]);
-
-		// Override the default toString on arrays so that we get a comma-delimited list
-		// TODO: Implement toString on observable list?
-		// val.toString = Property$_arrayToString.bind(val);
-	}
-
-	return val;
-}
-
 export function Property$init(property: Property, obj: Entity, val: any): void {
 	Property$pendingInit(obj, property, false);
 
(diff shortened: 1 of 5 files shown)


src/property.ts
```diff
@@ -913,26 +910,6 @@ function Property$subArrayEvents(obj: Entity, property: Property, array: Observa
 	});
 }
 
-function Property$getInitialValue(property: Property, obj: Entity): any {
```
mattheyan, line 916: Moved into `Property$init`

src/property.ts
```diff
@@ -233,10 +233,7 @@ export class Property implements PropertyPath {
 				else
 					throw new Error(`Invalid property 'init' option of type '${getTypeName(options.init)}'.`);
 
-				const property = this;
-				this.initializer = function () {
-					return targetType.model.serializer.deserialize(this, initFn.call(this), property, new InitializationContext(true));
-				};
+				this.initializer = initFn;
```
mattheyan, line 236: Store raw initializer function (result will be handled / transformed in `Property$init`)

src/property.ts
```diff
@@ -953,7 +930,37 @@ function Property$ensureInited(property: Property, obj: Entity): void {
...
+			if (property.isConstant)
+				initValue = typeof property.constant === "function" ? (property.constant = property.constant()) : property.constant;
+			else if (property.initializer) {
+				initValue = property.initializer.call(obj);
+				if (isEntityType(property.propertyType)) {
+					initData = initValue;
+					if (property.isList)
+						initValue = [];
+					else
+						initValue = new property.propertyType();
```
mattheyan, line 946: When `init` is used for a reference property, establish the property value (array or new object) prior to using the actual data, since it could cause rules to run which could impact other objects.

src/property.ts
```diff
@@ -953,7 +930,37 @@ function Property$ensureInited(property: Property, obj: Entity): void {
...
+
+				// Override the default toString on arrays so that we get a comma-delimited list
+				// TODO: Implement toString on observable list?
+				// val.toString = Property$_arrayToString.bind(val);
+			}
+
+			Property$init(property, obj, initValue);
+
+			if (initData)
+				obj.update(property.name, initData);
```
mattheyan, line 963: When using the `init` option, after setting the property in the model, call `update` to initialize the object (ex: if the `init` function returned JSON).

src/property.ts
```diff
@@ -913,14 +910,12 @@ function Property$subArrayEvents(obj: Entity, property: Property, array: Observa
...
 
-function Property$getInitialValue(property: Property, obj: Entity): any {
+function Property$getInitialValue(property: Property): any {
 	// Constant
 	if (property.isConstant)
 		return typeof property.constant === "function" ? (property.constant = property.constant()) : property.constant;
 
-	var val = property.initializer
-		? property.initializer.call(obj)
-		: property.defaultValue;
```
mattheyan, line 923: Don't call the initializer when getting the initial (default) value for a property, since a context-aware function could cause object creation, rule execution, etc. which could result in infinite "recursion" in the case of bi-directional relationships between objects and properties. Instead, the property will be given a default value, and then the initializer will be used to set or update the value (in `Property$ensureInited`).

src/property.ts
```diff
@@ -953,7 +948,9 @@ function Property$ensureInited(property: Property, obj: Entity): void {
 
 		// Do not initialize calculated properties. Calculated properties should be initialized using a property get rule.
 		if (!property.isCalculated) {
-			Property$init(property, obj, Property$getInitialValue(property, obj));
+			Property$init(property, obj, Property$getInitialValue(property));
+			if (property.initializer)
+				obj.update(property.name, property.initializer.call(obj));
```
mattheyan, line 953: After initializing the property with its default value, use the initializer function if provided to set or update the property.

tskimmett (approved):

PR #112, opened 2023-01-26
Fix error "Cannot read properties of null (reading 'ready')" when using async resolver and initializers

mattheyan:


test/initialize-with-async-resolver.unit.ts
```diff
@@ -0,0 +1,80 @@
...
+			}));
+		});
+
+		it("can be established via property initializers", async () => {
+			const consoleWarn = jest.spyOn(console, "warn").mockImplementation(() => {});
+			const instance = await model.types.Root.create({
+				"State": "SC",
+				"OuterSection": {}
+			}) as any;
+			expect(consoleWarn).not.toBeCalled();
```
mattheyan, line 75: New test that demonstrates the issue

test/back-references.unit.ts
```diff
@@ -1,88 +1,12 @@
...
-import { Property } from "../src/property";
-import { isEntityType } from "../src/type";
+import { ensureChildProperties } from "./utils";
+import { IgnorePropertyConverter } from "./ignore-property-converter";
+import { InitializeBackReferencesConverter } from "./initialize-back-references-converter";
+import { IdReferencePropertyConverter } from "./id-reference-property-converter";
 
 require("../src/resource-en");
 
-class IgnorePropertyConverter extends PropertyConverter {
```
mattheyan, line 10: Moved this stuff to separate files so that they could be reused across test modules.

src/entity.ts
```diff
@@ -212,19 +212,21 @@ export class Entity {
 			return context.ready.then(markPersistedWhenIdAssigned);
 		}
 
+		const context = this._context;
```
mattheyan, line 215: fix: Capture the context that will be used when the method returns

tskimmett (approved):

PR #118, opened 2023-05-15
Don't ignore null value in state for entity type properties

### LINKED ISSUE #115: Cannot initialize a reference property to null
If an object's initial state for a reference property is `null`, the value is ignored by the serializer.

See [entity-serializer.ts](https://github.com/cognitoforms/model.js/blob/805be3e71e164e790b53fa8445a4438968df7b7c/src/entity-serializer.ts#LL218C5-L218C5)

Most likely the property still is initialized to `null` by default, since it is the data type default. But, this is dependent on the property configuration and conceivable would not happen in some cases. It also results in the property being set to "pending init", which should not be the case since the object was explicitly provided an initial value of `null` for the property.

mattheyan:
Fixes #115 - Cannot initialize a reference property to null

> Most likely the property still is initialized to null by default, since it is the data type default. But, this is dependent on the property configuration and conceivable would not happen in some cases. It also results in the property being set to "pending init", which should not be the case since the object was explicitly provided an initial value of null for the property.

### DIFF
--- a/src/entity-serializer.ts
+++ b/src/entity-serializer.ts
@@ -216,6 +216,8 @@ export class EntitySerializer {
 				value = data;
 			else if (data instanceof Object)
 				value = resolveEntity(ChildEntity, data);
+			else
+				value = data;
 		}
 
 		// Value List
--- a/src/entity.unit.ts
+++ b/src/entity.unit.ts
@@ -4,6 +4,7 @@ import { Entity, EntityConstructorForType, isEntity } from "./entity";
 import "./resource-en";
 import { CultureInfo } from "./globalization";
 import { updateArray } from "./observable-array";
+import { Property$pendingInit } from "./property";
 
 let Types: { [name: string]: EntityConstructorForType<Entity> };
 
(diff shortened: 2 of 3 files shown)


src/entity-serializer.ts
```diff
@@ -216,6 +216,8 @@ export class EntitySerializer {
 				value = data;
 			else if (data instanceof Object)
 				value = resolveEntity(ChildEntity, data);
+			else
+				value = data;
```
mattheyan, line 220: fix: Don't ignore a `null` value (or really any other type of value)

src/entity.unit.ts
```diff
@@ -163,6 +164,21 @@ describe("Entity", () => {
...
+			const movieProp = Types.Person.meta.getProperty("Movie");
+			expect(Property$pendingInit(person, movieProp)).toBe(true);
+			expect(person.Movie).toBe(null);
+			// expect(Property$pendingInit(person, movieProp)).toBe(false);
+		});
+
+		it("can initialize reference properties to null", () => {
+			const person = Types.Person.meta.createSync({ Id: "1", Movie: null });
+			const movieProp = Types.Person.meta.getProperty("Movie");
+			expect(Property$pendingInit(person, movieProp)).toBe(false);
```
mattheyan, line 178: Test: Ensure that `pendingInit` is `false` after initializing the property to `null`

gsmcdonald (approved):

PR #120, opened 2023-05-19
Fixes for pending init logic

mattheyan:
- Get rid of special-case for default value rules and setting "pending init" when ensuring a list property is initialized
  - Don't run on initNew and always set pending init
- Always set pendingInit = true when ensureInited is called, except for contant properties and those with an initializer
- Always set pendingInit = false when a property is set or when initial state is provided


src/property.ts
```diff
@@ -1017,9 +1020,16 @@ function Property$setValue(property: Property, obj: Entity, currentValue: any, n
...
+
+		let updates = 0;
 		currentArray.batchUpdate((array) => {
-			updateArray(array, newValue);
+			updates = updateArray(array, newValue, true).length;
 		}, additionalArgs);
+
+		// If there were no updates (ex: array was previously an empty array and newValue is also an empty array),
+		// then set pendingInit to false here, since the array change event handler would not have fired
+		if (updates === 0)
```
mattheyan, line 1031: Simplify to always set pending init to false

src/observable-array.ts
```diff
@@ -608,8 +608,9 @@ function observableSplice(arr: any[], events: any[], index: number, removeCount:
 	}
 }
 
-export function updateArray(array: any[], values: any[] /*, trackEvents */): any[] {
-	var trackEvents: boolean = arguments[2];
+export function updateArray(array: any[], values: any[]): void;
+export function updateArray(array: any[], values: any[], trackEvents: true): any[];
+export function updateArray(array: any[], values: any[], trackEvents: boolean = false): any[] | void {
```
mattheyan, line 613: Is trackEvents still needed / used?

mattheyan: Removed

src/calculated-property-rule.ts
```diff
@@ -33,9 +34,6 @@ export class CalculatedPropertyRule extends Rule {
 			if (options.property) {
 				property = typeof options.property === "string" ? rootType.getProperty(options.property) as Property : options.property as Property;
 
-				if (options.isDefaultValue && property.isList)
-					(options as RuleInvocationOptions).onInitNew = true;
```
mattheyan, line 37: No longer special-case default rules to run on init

src/entity.ts
```diff
@@ -300,8 +303,12 @@ export class Entity {
 					value = Entity.createOrUpdate(ChildEntity.meta, state, this._context);
 			}
 		}
-		else if (prop.isList && Array.isArray(state) && Array.isArray(currentValue))
+		else if (prop.isList && Array.isArray(state) && Array.isArray(currentValue)) {
 			currentValue.splice(0, currentValue.length, ...state.map(s => this.serializer.deserialize(this, s, prop, this._context)));
+
+			// Mark property pendingInit = false, since the setter is not called
+			Property$pendingInit(this, prop, false);
+		}
```
mattheyan, line 311: Ensure that pending init is set to false when a property has initial state provided when the object is constructed

src/helpers.ts
```diff
@@ -329,3 +329,43 @@ export function entries(obj: any): [string, any][] {
...
 }
+
+function callSplice(arr: any[], index: number, removeCount: number, addItems: any[]): void {
+	if (removeCount || addItems.length > 0) {
+		var addItemsArgs = [index, removeCount].concat(addItems);
+		arr.splice.apply(arr, addItemsArgs);
+	}
+}
+
+export function updateArray(array: any[], values: any[]): void {
```
mattheyan, line 340: Removed the unused event tracking code from `updateArray` and moved it from the "observable-array" file to "helpers" since it no longer contains any "observable" specific code.

src/observable-array.ts
```diff
@@ -563,84 +563,3 @@ export class ArrayObserver<ItemType> {
...
-				oldStartingIndex: null,
-				oldItems: null,
-				newStartingIndex: index,
-				newItems: addItems
-			});
-		}
-	}
-}
-
-export function updateArray(array: any[], values: any[] /*, trackEvents */): any[] {
```
mattheyan, line 611: Moved to "helpers"

src/property.ts
```diff
@@ -896,6 +896,8 @@ export function Property$pendingInit(obj: Entity, prop: Property, value: boolean
 
 function Property$subArrayEvents(obj: Entity, property: Property, array: ObservableArray<any>): void {
 	array.changed.subscribe(function (args) {
+		Property$pendingInit(obj, property, false);
```
mattheyan, line 899: Set pending init to false when an array is directly modified (ex: calling a method like `splice`)

src/property.ts
```diff
@@ -950,14 +952,16 @@ function Property$ensureInited(property: Property, obj: Entity): void {
...
-			const underlyingValue = obj.__fields__[property.name];
-			// Mark the property as pending initialization if it still has no underlying value, or is the property type's default, to allow default calculation rules to run for it
-			// List properties are defaulted onInitNew instead of on access
-			if (underlyingValue === property.defaultValue && !Array.isArray(underlyingValue))
+				setPendingInit = false;
+			}
+			// Mark the property as pending initialization if the property value may need to
+			// be established by a default calculation rule, or some other external logic.
+			// This is not relevant if it is a constant or its value is established via an initilizer.
+			if (setPendingInit)
```
mattheyan, line 964: Set pending init to true not based on the value, but whether or not it was initialized with an initializer.

Also, constants should not be pending init.

src/property.ts
```diff
@@ -1017,9 +1021,14 @@ function Property$setValue(property: Property, obj: Entity, currentValue: any, n
...
 	if (property.isList) {
 		let currentArray = currentValue as ObservableArray<any>;
+
 		currentArray.batchUpdate((array) => {
 			updateArray(array, newValue);
 		}, additionalArgs);
+
+		// Set pendingInit to false here, since an array change event will not be raised if there are no changes
+		// (ex: array was previously an empty array and newValue is also an empty array).
+		Property$pendingInit(obj, property, false);
```
mattheyan, line 1031: Set pending init to false when a list property is set, regardless of whether it results in a change to the list

src/property.unit.ts
```diff
@@ -435,4 +458,306 @@ describe("Property", () => {
...
+				expect(Property$pendingInit(instance, property)).toBe(false);
+			});
+
+			it("is true for an existing object without a value", async () => {
+				const Person = valuePropModel.types.Person;
+				const instance = await Person.create({ Id: "4" }) as any;
+				const property = Person.getProperty("Name");
+				expect(instance.Name).toBeNull();
+				expect(Property$pendingInit(instance, property)).toBe(true);
+			});
```
Christopher-Masselli, line 510: _Test suggestion_ an undefined/null string or a Boolean that's starts as false

src/entity.ts
```diff
@@ -271,6 +271,9 @@ export class Entity {
 					// Remove excess items from the list
 					currentValue.splice(state.length);
 				}
+
+				// Mark property pendingInit = false, since the setter is not called
```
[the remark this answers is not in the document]
mattheyan: Its saying that if the setter was called it would set pending init to false, but since its not called we have to set pending init here.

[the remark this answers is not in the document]
mattheyan: If the `value` variable is set, then it calls the setter at the bottom of this method, and the setter will set pending init = false. In some cases value is not assigned and instead the current value (i.e. array) is directly manipulated. If this resulted in changes to the array then the array change handler would set pending init. But, if there were no changes to the array then it wouldn't, so that's why I set it here as well.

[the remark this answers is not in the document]
mattheyan: Honestly it probably _should_ just call the setter. But, I'd rather not pull that into the scope of this PR. 😬

src/property.unit.ts
mattheyan: The same test but with a boolean value? I'm guessing you're suggesting this because this ensures that the pending init logic isn't affected by data types?

mattheyan: This actually uncovered another edge case. When initial state is passed in for a new object the setter is called but if the value isn't changing it doesn't call `setValue`, so pending init wasn't getting set to false. I don't think this impacted product functionality though.

src/entity.ts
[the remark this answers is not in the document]
mattheyan: Yeah, I think that's the idea in theory, that its going to take the given raw state and set the property. In many cases it does just that (calls the setter), but list properties are just a bit of an edge case since it directly modifies the list instead. It may be that it was done this way because until recently the setter would throw an error if you called it for a list property.

src/observable-array.ts
mattheyan: Reverted

PR #125, opened 2023-09-27
TypeScript type enhancements and test fixes

mattheyan:
**Summary**: model.js provides an entity type system, but does not effectively leverage TypeScript. This PR introduces a `createModel` function that enables creating a model that is type aware, and generic types that allow the model.js APIs to leverage the specific types of the model.

_Example_:
```
type Person = { Name: string };
type Namespace = { Person: Person };

// The createModel arguments are type checked based on the types in the namespace (i.e. Person)
const model = await createModel<Namespace>({
    Person: {
        Name: String,
        Initials: {
            get() {
                // "this" is EntityOfType<Person>
            }
        }
    }
});

// The model knows what types it contains
const personType = model.Person;

// The types included on the model are TS type-specific, and type check arguments and return specific types
const person = Person.meta.create({ Name: "Alice" });

// The resulting entity is `EntityOfType<Person>`
console.log(person.Name);
```

### DIFF
--- a/src/condition-rule.ts
+++ b/src/condition-rule.ts
@@ -10,7 +10,7 @@ export class ConditionRule extends Rule {
 	assert: (this: Entity) => boolean;
 
 	// The message describing the state of the condition
-	message: string | ((this: Entity) => string);
+	message: string | ((this: Entity) => string | null | undefined);
 
 	// Array of property paths the validation condition should be attached to when asserted, in addition to the target property
 	properties: PropertyPath[];
@@ -83,7 +83,7 @@ export interface ConditionRuleOptions extends RuleOptions {
 	assert?: (this: Entity) => boolean;
 
 	// the message to show the user when the validation fails
-	message?: string | ((this: Entity) => string);
+	message?: string | ((this: Entity) => string | null | undefined);
 
 	// an array of property paths the validation condition should be attached to when asserted, in addition to the target property
 	properties?: PropertyPath[];
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -384,7 +384,7 @@ export class Entity {
 	}
 }
 
-export type EntityPropertiesOfType<T> = {
+export type EntityDynamicPropertiesOfType<T> = {
     [P in keyof T]:
 		T[P] extends (infer TItem)[]
 			? ObservableArray<(
@@ -410,7 +410,7 @@ export interface TEntity<T> extends Entity {
 	readonly changed: EventSubscriber<Entity, EntityChangeEventArgsForType<T>>;
 }
 
-export type EntityOfType<T> = TEntity<T> & EntityPropertiesOfType<T>;
+export type EntityOfType<T> = TEntity<T> & EntityDynamicPropertiesOfType<T>;
 
 export type EntityConstructorForType<T> = {
     new(id: string, args?: EntityArgsOfType<T>): EntityOfType<T>;
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -5,7 +5,7 @@ import { InitializationContext } from "./initilization-context";
 import { ObjectMeta, ObjectMetaOfType } from "./object-meta";
 import { Property, Property$init, Property$pendingInit, Property$setter } from "./property";
 import { ObjectLookup, entries } from "./helpers";
-import { DefaultSerializationSettings } from "./entity-serializer";
+import { DefaultSerializationSettings, SerializationSettings } from "./entity-serializer";
 import { ObservableArray } from "./observable-array";
 
 export class Entity {
@@ -353,7 +353,7 @@ export class Entity {
 	 * Produces a JSON-valid object representation of the entity.
 	 * @param entity
 	 */
-	serialize(settings = DefaultSerializationSettings): object {
+	serialize(settings = DefaultSerializationSettings): Record<string, any> {
 		return this.serializer.serialize(this, settings);
 	}
 
@@ -407,6 +407,7 @@ export interface EntityBasePropertiesOfType<T> {
 	meta: ObjectMetaOfType<T>;
 	update(property: string, value?: any): Promise<void>;
 	update(args: EntityArgsOfType<T>): Promise<void>;
+	serialize(settings?: SerializationSettings): T;
 	readonly accessed: EventSubscriber<Entity, EntityAccessEventArgsForType<T>>;
 	readonly changed: EventSubscriber<Entity, EntityChangeEventArgsForType<T>>;
 }
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -188,6 +188,7 @@ export class Entity {
 	}
 
 	update(properties: ObjectLookup<any>): Promise<void>;
+	update(property: string, value: any): Promise<void>;
 	update(property: any, value?: any): Promise<void> {
 		let properties: ObjectLookup<any>;
 
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -2,7 +2,7 @@ import { Event, EventObject, EventSubscriber } from "./events";
 import { Format } from "./format";
 import { Type, EntityType, isEntityType, getIdFromState, TypeOfType } from "./type";
 import { InitializationContext } from "./initilization-context";
-import { ObjectMeta } from "./object-meta";
+import { ObjectMeta, ObjectMetaOfType } from "./object-meta";
 import { Property, Property$init, Property$pendingInit, Property$setter } from "./property";
 import { ObjectLookup, entries } from "./helpers";
 import { DefaultSerializationSettings } from "./entity-serializer";
@@ -187,7 +187,6 @@ export class Entity {
 	}
 
 	update(properties: ObjectLookup<any>): Promise<void>;
-	update(property: string, value: any): Promise<void>;
 	update(property: any, value?: any): Promise<void> {
 		let properties: ObjectLookup<any>;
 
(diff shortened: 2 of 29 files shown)


src/entity.ts
```diff
@@ -383,49 +384,101 @@ export class Entity {
...
+			) | null;
+};
+
+type EntityOfTypeMemberOverrides<T> = {
+    update(args: EntityArgsOfType<T>): Promise<void>;
+};
+
+export type EntityOfType<T> = EntityPropertiesOfType<T> & EntityOfTypeMemberOverrides<T> & Entity; // Omit<Entity, keyof EntityOfTypeMemberOverrides<T>> & ;
+
+export type TEntityConstructor<T> = {
```
mattheyan, line 412: TODO: Merge with EntityConstructorForType

src/format.ts
```diff
@@ -82,8 +82,9 @@ export abstract class Format<T> {
 		return new CustomFormat<T>(model, options);
 	}
 
-	static fromTemplate<TEntity extends Entity>(type: Type, template: string, formatEval?: (tokenValue: string) => string): Format<TEntity> {
-		return new ModelFormat<TEntity>(type, template, formatEval);
+	// TODO: Infer from type? (Doesn't provide any value in this context)
+	static fromTemplate<EntityType extends Entity>(type: Type, template: string, formatEval?: (tokenValue: string) => string): Format<EntityType> {
```
mattheyan, line 86: TODO: Infer from type? (Doesn't provide any value in this context)

src/property.ts
```diff
@@ -394,7 +394,7 @@ export class Property implements PropertyPath {
 
 				targetType.model.ready(() => {
 					let onChangeOf: PropertyPath[] = resolveDependsOn(this, "length", options.length.dependsOn);
-					if (isEntityType(this.propertyType)) {
+					if (this.isList) {
```
mattheyan, line 397: _FIX_: This is a functional change. For some reason we were treating all "length" rules on entity-typed properties as list-length and string-length otherwise (assuming it was either `Entity[]` or `string`). So, you couldn't create a list-length rule on a property of type `string[]`. There was a test that was getting around this by declaring a property to be an entity list property, but actually using string values in reality.

src/condition-rule.ts
```diff
@@ -10,7 +10,7 @@ export class ConditionRule extends Rule {
 	assert: (this: Entity) => boolean;
 
 	// The message describing the state of the condition
-	message: string | ((this: Entity) => string);
+	message: string | ((this: Entity) => string | null | undefined);
```
mattheyan, line 13: _info_: Allow `message` and `error` functions to return `null` or `undefined` to avoid TS errors, and the rules should handle null/undefined anyway.

src/entity.ts
```diff
@@ -383,49 +384,101 @@ export class Entity {
 	}
 }
 
+export type EntityPropertiesOfType<T> = {
```
mattheyan, line 387: _info_: Add `EntityPropertiesOfType<T>` type, where `T` is the "schema" of the type (the properties it has and their TS types. Lists are defined as non-nullable `ObservableArray<T>` and non-lists are defined as `T | null`.

src/entity.ts
```diff
@@ -383,49 +384,101 @@ export class Entity {
...
+
+export type EntityOfType<T> = EntityPropertiesOfType<T> & EntityOfTypeMemberOverrides<T> & Entity; // Omit<Entity, keyof EntityOfTypeMemberOverrides<T>> & ;
+
+export type TEntityConstructor<T> = {
+    new(id: string, args?: EntityArgsOfType<T>): EntityOfType<T>;
+    new(args?: EntityArgsOfType<T>): EntityOfType<T>;
+	meta: TypeOfType<T>;
+};
+
+export type EntityArgsOfType<T> = Partial<{
```
mattheyan, line 418: _info_: Define type `EntityArgsOfType<T>`, where `T` is the "schema" of the type (the properties it has and their TS types). This is intended to be used for constructors, `create` calls, `update` calls, etc. All properties are optional, so they can be omitted. Non-lists can be null, but lists should not be null. The supported value types are covered, and then anything else is assumed to be an entity property. Entity property values can be an instance of the entity or the supported "args" (i.e. this type) for that entity type.

src/entity.ts
mattheyan:
```ts
type Example = {
  text: string;
  values: number[];
  ref: OtherEntity;
  children: ChildEntity[];
};

type ExampleEntityProperties = EntityPropertiesOfType<Example> = {
  text: string | null;
  values: ObservableArray<number>;
  ref: OtherEntity | null;
  children: ObservableArray<ChildEntity>;
};
```

src/entity.ts
mattheyan:
```ts
type Example = {
  text: string;
  values: number[];
  ref: OtherEntity;
  children: ChildEntity[];
};

type ExampleEntityArgs = EntityArgsOfType<Example> = {
  text?: string | null;
  values?: number[];
  ref?: EntityOfType<OtherEntity> | EntityArgsOfType<OtherEntity> | null;
  children?: (EntityOfType<ChildEntity> | EntityArgsOfType<ChildEntity>)[];
};
```

src/entity.ts
mattheyan: Also, `readonly` properties are excluded, since they would represent calculated properties that can't be set.

src/entity.ts
mattheyan: Also, `readonly` properties are excluded, since they would represent calculated properties that can't be set.

src/entity.ts
```diff
@@ -383,49 +384,101 @@ export class Entity {
...
+// TODO: Replace with TEntityConstructor, rename to EntityConstructorOfType
+export interface EntityConstructorForType<EntityType extends Entity> extends EntityConstructor {
+	new(): EntityType;
+	new(properties?: ObjectLookup<any>): EntityType; // Construct new instance with state
 	meta: Type;
 }
 
-export interface EntityRegisteredEventArgs {
-	entity: Entity;
+export interface EntityRegisteredEventArgs<EntityType extends Entity> {
```
mattheyan, line 450: _info_: Add generic type argument `EntityType` to a lot of types so that they can leverage more specific typing when needed.

src/type.ts
```diff
@@ -504,29 +504,46 @@ export class Type {
 
 export type Value = string | number | Date | boolean;
 export type ValueType = StringConstructor | NumberConstructor | DateConstructor | BooleanConstructor;
-export type EntityType = EntityConstructorForType<Entity>;
-export type PropertyType = ValueType | EntityType | ObjectConstructor;
+export type ReferenceType = EntityConstructorForType<Entity>;
```
mattheyan, line 507: _info_: Rename `EntityType` to `ReferenceType` to avoid confusion with `EntityType` generic type arguments, and also is more consistent with reference/value terminology used in ExoModel.

mattheyan: Might want to revert this...

mattheyan: Reverted

src/property.ts
```diff
@@ -955,7 +956,10 @@ function Property$ensureInited(property: Property, obj: Entity): void {
 			let setPendingInit = !property.isConstant;
 			Property$init(property, obj, Property$getInitialValue(property));
 			if (property.initializer) {
-				obj.update(property.name, property.initializer.call(obj));
+				const args: ObjectLookup<any> = {};
+				args[property.name] = property.initializer.call(obj);
+				// TODO: Could this just be obj[property.name] = ...?
+				obj.update(args);
```
mattheyan, line 962: TODO: Type checking error on `obj.update`

src/model.ts
```diff
@@ -370,15 +370,34 @@ export class Model {
...
+export interface ModelOfType<TTypes> extends Model {
+	readonly types: { [T in keyof TTypes]: TypeOfType<TTypes[T]> };
+}
+
+export interface ModelWithNamespace<TNamespace> extends Model {
+	readonly $namespace: ModelNamespace<TNamespace>;
 }
 
-export type ModelOptions = {
+export interface ModelConstructor<TNamespace> {
```
mattheyan, line 392: Is this used? Also, `TNamespace`.

src/event-scope.unit.ts
```diff
@@ -160,7 +179,6 @@ describe("EventScope", () => {
 					type: Boolean,
 					get: {
 						dependsOn: "{FullName,FirstName,LastName}",
-						defaultIfError: undefined, // This is necessary to make sure the rule will fail if it throws an error
```
mattheyan, line 163: _info_: This did not appear to be used by the property extend code

src/model.ts
```diff
@@ -370,15 +370,34 @@ export class Model {
 	}
 }
 
-export interface ModelConstructor {
-	new(createOwnProperties?: boolean): Model;
+export function createModel<TTypes>(options?: ModelTypeOptions<TTypes> & Required<ModelNamespaceOption<TTypes>> & ModelLocalizationOptions, config?: ModelConfiguration): Promise<ModelOfType<TTypes> & ModelWithNamespace<TTypes>>;
+export function createModel<TTypes>(options?: ModelTypeOptions<TTypes> & ModelLocalizationOptions, config?: ModelConfiguration): Promise<ModelOfType<TTypes> & ModelNamespace<TTypes>>;
+export function createModel<TTypes>(options?: ModelOptions<TTypes>, config?: ModelConfiguration): Promise<ModelOfType<TTypes>> {
```
mattheyan, line 375: _info_: Add a `createModel` function that can be used to provide types for the model and returns a model that knows what types it contains.

tskimmett (approved):

